PR #725, opened 2014-05-20
Enyo-4039: partial fix for enyo.load()

greenITea:
fix for loading of multiple files/packages at once on dom loading time:
enyo.load(["a","b","c"]);


webOS101:
Thank you for these contributions.  The new 2.7.0 version of Enyo does not use enyo.load anymore so we cannot bring these commits with us to 2.7.  You probably (hopefully!) can use the new `request()` feature to dynamically load modules at runtime.

PR #726, opened 2014-05-20
Enyo 4037

greenITea:
ENYO-4037: fix not working enyo.load() after a loading error


webOS101:
Thank you for these contributions. Enyo 2.7.0 uses a new tool for packaging apps so this code no longer applies.

PR #730, opened 2014-05-22
ENYO-4038: fix nodejs command in minify.sh

greenITea:
add better check for nodejs in minify.sh


dmikey:
If node doesn't work, this is most likely a linking problem with your environment. Can you provide more information about your system? OS specifically.

webOS101:
I believe some installations do install node as nodejs as opposed to node.  If we decide we want to do this (and think we had something like this before), we'll need to merge into other places that call node (bootplates, sampler).

dmikey:
Isn't that as easy as tossing a soft link into the bin? 

I'm still leaning on the side of this being an environment setup need. Usually in a shell script you also reference the node link vs nodejs.

#!/usr/bin/env node

greenITea:
I use ubuntu. In ubuntu nodejs is installed in path /usr/bin/nodejs. The reason may be an other application, which name is node. There is an nodejs-legacy package awailable. It just creates a symlink to nodejs and solves the problem. But you need to know that.

dmikey:
I'm going to close this, because I still feel strongly that this is an environment setup issue, this wouldn't alleviate the problem with other NodeJS interactions on Ubuntu, and I would hate to compensate for something that should be done on the environment.

PR #739, opened 2014-05-25
fix updateProgress in ie8

johannjacobsohn:
"event" is probably meant to be "ev"; IE8 throws an exception otherwise.

Enyo-DCO-1.1-Signed-off-by: Johann Jacobsohn j.jacobsohn@satzmedia.de

### DIFF
--- a/source/ajax/Ajax.js
+++ b/source/ajax/Ajax.js
@@ -251,7 +251,7 @@ enyo.kind({
 				ev[k] = event[k];
 			}
 		}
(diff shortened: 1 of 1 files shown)


ryanjduffy:
Since IE8 doesn't properly support the onprogress event (it doesn't provide any progress details), I've opened another PR to guard updateProgress in that scenario. Feel free to comment here or there if you have any concerns.

https://github.com/enyojs/enyo/pull/995

PR #740, opened 2014-05-28
changes to getTouchStrategy

rafabernad:
Enable TranslateScrollStrategy on Android Kitkat and later, and TransitionScrollStrategy on iOS5 and later.


ryanjduffy:
This PR is partially superceded by #835. Regarding TransitionScrollStrategy, it hasn't received the necessary testing to make it the default strategy for > iOS 5. For the time being, it's considered to be 'use at your own risk.'

PR #750, opened 2014-06-05
ENYO-4048: Allow for overriding of scroller options with falsy values.

aarontam:
Issue
-------
It is not currently possible to override scroller properties with falsy values (i.e. properties that are true by default, such as `useMouseWheel`).

Fix
----
Removed the `exists` option when mixing in `scrollerOptions` as there did not appear to be a current need for this (and there is no relevant history for the existence of this option here).


lishammel:
@aarontam  - do we still need this PR merged? thanks.

aarontam:
Hi @lishammel I believe we still need this but it had fallen off the radar. To be fair, it doesn't seem like the use case that exposed this issue is very common, but is still relevant.

KunmyonChoi:
@aarontam Nice expression "fallen off the radar" :)

lishammel:

PR #803, opened 2014-07-18
Bhv 5137 Add renderOnShow option

DavidUM:
Requested from HQ developers.

Issue
------
Due to performance matter, app developers would like to adjust rendering timing between render on creating and render on actual showing.

Cause
-------
enyo.Control does not have an option to delay rendering

Fix
----
Add renderOnShow option. If developer sets it true, 
automatically this control has showing:false and canGenerate: false
When we call setShowing(true) first time, it will start render.

### DIFF
--- a/source/dom/Control.js
+++ b/source/dom/Control.js
@@ -143,7 +143,11 @@
 		*/
 		renderOnShow: false,
 		/**
-		* private
+		* When you make a control which renderOnShow is true visible,
+		* you should know where it placed.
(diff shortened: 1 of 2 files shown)


lishammel:
@DavidUM  and @aarontam  - is this PR still needed? or is it replaced by the work Aaron has been doing? thanks.

aarontam:
Hi @lishammel, I think we can close this for now, thanks!

PR #872, opened 2014-10-01
Fix for: BHV-5137

clinuz:
# Task

Add a `renderOnShow` option to enyo.Control as a base API that allows individual controls (and their children) to be created but not rendered until explicitly set via the `showing` property.

Enyo-DCO-1.1-Signed-Off-By: Cole Davis (cole.davis@lge.com)


lishammel:
@clinuz  and @aarontam  - should this PR be closed out?

aarontam:
Hi @lishammel, yep I believe we can close this for now.

PR #890, opened 2014-10-18
fix for exceptions seen coming from transition scroller

MichaelRando:


lishammel:
@MichaelRando  - would love some additional details on this.  Thanks!

lishammel:
@MichaelRando  - let us know if you can provide additional details. Thanks!

MichaelRando:
How can I help?

lishammel:
Hi @MichaelRando  - per @dmikey  if you could give some more detail on when and where you see the scroller throwing exceptions that would be helpful for us to use to verify the fix. Thanks!

MichaelRando:
It's a simple fix, I would imagine you could verify the fix by inspection, considering that you're splitting a value that has a possible equivalence to string 'none' and all my check did was not split it in that case. 
I think you could test it out if you called the old code redundantly, that 'none' string is right there.

webOS101:
Michael, thanks for this pull request.  After some internal review we've decided not to take this pull request as we push out 2.7.  If you experience any of these issues with 2.7 we can look at new pull requests in the context of 2.7.

PR #922, opened 2014-11-08
ENYO-442: Add a method to temporarily suppress events for a given set of code.

aarontam:
### Issue
There were several instances where we needed to surround code with a pair of `silence`/`unsilence` calls to temporarily suppress event firing.

### Fix
We've implemented a `muzzle` method (I'm not married to the choice of name) that takes either a string method name, or a function, and surrounds its execution with a `silence` and corresponding `unsilence` call.

### DIFF
--- a/source/kernel/Component.js
+++ b/source/kernel/Component.js
@@ -903,6 +903,42 @@
 		},
 
 		/**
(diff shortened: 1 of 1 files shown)


lishammel:
Closing since Jira has been closed and was decided that these changes are not needed as of now.

PR #960, opened 2014-12-10
Fixes ENYO-527

ryanjduffy:
## Issue
Need to provide API to determine which controls are visible for a DataList and DataGridList.

## Fix
Add `getVisibleControlRange()` to enyo.DataList which returns the `start` and `end` index of the visible controls. For a control to be considered visible, a certain percentage (established by the `visibleThreshold` property) must be visible.

### DIFF
--- a/source/ui/data/VerticalDelegate.js
+++ b/source/ui/data/VerticalDelegate.js
@@ -778,7 +778,7 @@
 					end: -1
 				},
 				posProp = list.posProp,
-				sizeProp  = list.psizeProp,
+				sizeProp = list.psizeProp,
 				size = this[sizeProp](list),
 				scrollPosition = this.getScrollPosition(list),
 				pages = list.pages.slice(0).sort(function (a, b) {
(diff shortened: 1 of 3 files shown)


seunghoh:
I can see the problem on garnet is fixed. It works well on garnet DataList

source/ui/data/VerticalDelegate.js
```diff
@@ -128,6 +128,8 @@
 			// get our initial sizing cached now since we should actually have
 			// bounds at this point
 			this.updateBounds(list);
+			// calc offset of pages to scroller client
+			this.calcScrollOffset(list);
```
aarontam: Do we need to add this call to VerticalGridDelegate's `rendered` as well?

ryanjduffy: yes

ryanjduffy: Done!

aarontam: :+1:

PR #988, opened 2015-01-02
ENYO-733: Default to building resolution-independent CSS.

aarontam:
### Issue
We are currently making resolution-independence opt-in via the deploy process, which requires the specification of the `-r` or `--ri` flag, requiring the potential addition of additional switching logic at the build level.

### Fix
Because we have validated our earlier work and we know that we always want resolution-independence in `2.5-gordon`, we default the compilation of resolution-independent CSS to true and make this feature opt-out. This includes changing the flag to `-d` (dependence), which is the shorthand flag for `--rd` (resolution-dependence). Additionally, we make updates to consistently use the shorthand flags where appropriate.

### DIFF
--- a/tools/deploy.js
+++ b/tools/deploy.js
@@ -105,7 +105,7 @@ process.on('message', function(msg) {
 var node = process.argv[0],
 	deploy = process.argv[1],
 	less = true, // LESS compilation, turned on by default
-	ri = false, // LESS resolution-independence conversion, turned off by default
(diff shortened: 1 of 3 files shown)


graynorton:
Closing, at least for now. We probably don't want to make ri the default, except in the case of Moonstone apps for the webOS TV, so better to handle that in bootplate-moonstone and the webOS build recipes.

PR #994, opened 2015-01-08
ENYO-772: Allow specification of measurement precision.

aarontam:
### Issue
We are not currently limiting the number of decimal places for our converted resolution-independent measurements, resulting in long floating point numbers.

### Fix
A new property `precision` has been added to the resolution-independence plugin, that defaults to a value of `5`. Additionally, general clean-up and refactoring was done to improve semantics and specifically to separate out the logic for parsing a rule with a string value vs. a rule with separate value and unit objects. Lastly, some hardening was done to catch some other edge cases with unit parsing, and an unnecessary return statement was removed from `visitRule` as we are not specifying `isReplacing:true` in our plugin.


KunmyonChoi:
Hi, when base fontunit is 24, is the 1px equivalent rem value 0.04167? Which means minimum precision should be 5. How do you think?

aarontam:
Hi @KunmyonChoi that's a good point - let me update the default precision to reflect this.

PR #997, opened 2015-01-13
Fixes ENYO-785

ryanjduffy:
## Issue
Taps on controls within a popup that close the popup would be triggered on controls under the popup.

## Fix
Prevent taps from cascading to controls under a Popup by calling `event.preventDefault()`

### DIFF
--- a/source/ui/Popup.js
+++ b/source/ui/Popup.js
@@ -219,6 +219,7 @@
 		* @private
 		*/
 		handlers: {
+			tap: 'tap',
 			onkeydown: 'keydown',
 			ondragstart: 'dragstart',
 			onfocus: 'focus',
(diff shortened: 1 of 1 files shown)


source/ui/Popup.js
```diff
@@ -219,6 +219,7 @@
 		* @private
 		*/
 		handlers: {
+			tap: 'tap',
```
aarontam: I don't think we need to explicitly declare this as we call `tap` for tap events if a handler with this name exists. Alternatively, if we wanted to be explicit, then this should be `ontap`.

aarontam:
This works for the case of a button, but this general technique prevents an input from receiving focus.

ryanjduffy:
Bogus :cry:

ryanjduffy:
Trying again with a new PR

PR #1000, opened 2015-01-16
BHV-21381:InputSample: Removed unnecessary accumulation of directionality in style at the time of creation

rakesh25:
Issue:
In Input, directionality property in style is getting appended twice at the time of creation.

Fix:
No necessity of calling  "this.valueChanged" twice in create function of Input.js, as it is already been invoked from "this.placeholderChanged" function. Removed "this.valueChanged" call from create function.

Enyo-DCO-1.1-Signed-off-by: Rakesh kumar rakesh25.kumar@lge.com

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
     "nopt": "2.1.2"
   },
   "devDependencies": {
+    "chai": "~1.9.2",
     "sinon": "~1.8.1",
     "sinon-chai": "~2.5.0",
     "mocha": "~1.20.0",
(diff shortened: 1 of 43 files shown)


aarontam:
Hi @rakesh25 can you clean-up this PR and only include the requisite changes you've made instead of these other existing commits? Thanks!

rakesh25:
Closing this PR so to include only the requisite changes other than exisiting commits..Will create a new PR.

PR #1003, opened 2015-01-20
Fixes ENYO-785

ryanjduffy:
## Issue
Taps on controls within a popup that close the popup would be triggered on controls under the popup.

## Fix
Prevent taps from cascading to controls under a Popup by calling `event.preventDefault()`

### DIFF
--- a/source/ui/Popup.js
+++ b/source/ui/Popup.js
@@ -526,6 +526,19 @@
 		},
 
 		/**
+		* Prevent taps from cascading to controls covered by the popup
+		*
+		* @private
+		*/
+		tap: enyo.inherit(function (sup) {
+			return function (sender, event) {
+				if (!this.showing) {
+					event.preventDefault();
+				}
+			};
+		}),
+
+		/**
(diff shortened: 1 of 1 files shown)


source/ui/Popup.js
```diff
@@ -526,6 +526,19 @@
 		},
 
 		/**
+		* Prevent taps from cascading to controls covered by the popup
+		*
+		* @private
+		*/
+		tap: enyo.inherit(function (sup) {
```
clinuz: We overload the previous implementation but never call super, so overloading it (as opposed to replacing it) seems wrong. Also, enyo.Popup extends enyo.Control, and enyo.Control does not implement `tap` so this is a nop regardless. We don't return true (which may not be necessary because of the preventDefault call). Why do we only need to preventDefault if the popup isn't showing? If it wasn't showing, it wouldn't be receiving the tap? Also note the bogus and indirect way in which this handler is even called! enyo.Popup subverts the normal event system by arbitrarily registering (additionally) a handler `capturedTap` for tap events which allows this handler to get called anyway. Which happens first? How do we know that?

PR #1010, opened 2015-01-22
Fix: Android browser detection for some phones

wangii:
ZTE and Lenovo phones have browser userAgent like `Android/4.x.x`

Signed-off-by: Linan Wang<wang.linan at gmail dot com>

### DIFF
--- a/source/dom/platform.js
+++ b/source/dom/platform.js
@@ -45,6 +45,8 @@ enyo.platform = {
 		{platform: "androidChrome", regex: /Android .* Chrome\/(\d+)[.\d]+/},
 		// Android 2 - 4
 		{platform: "android", regex: /Android (\d+)/},
+		// Some android phones, e.g. ZTE and Lenovo
+		{platform: "android", regex: /Android\/(\d+)/},
 		// Kindle Fire
 		// Force version to 2, (desktop mode does not list android version)
 		{platform: "android", regex: /Silk\/1./, forceVersion: 2, extra: {silk: 1}},


lishammel:
Thanks for the PR - wanted to confirmed this is in the queue for review.

lishammel:
Just a quick update that initial review has finished. In the queue for second review and merge.

source/dom/platform.js
```diff
@@ -45,6 +45,8 @@ enyo.platform = {
 		{platform: "androidChrome", regex: /Android .* Chrome\/(\d+)[.\d]+/},
 		// Android 2 - 4
 		{platform: "android", regex: /Android (\d+)/},
+		// Some android phones, e.g. ZTE and Lenovo
+		{platform: "android", regex: /Android\/(\d+)/},
```
clinuz: Is it better to have a separate entry for "android" or simply modify the previous entry with something like `/Android(?:\/)?(\d+)/ `? Wouldn't that do the same thing?

clinuz:
If we get this updated to merge-able we can go ahead and move forward with it (and potentially update the line I commented on).

source/dom/platform.js
wangii: if it's common pattern i'd suggest to merge. keep rare/edge case distinct is good for maintaince, i believe.

aarontam: I think there are a fair number of devices with this user agent string pattern, where it would make sense to combine these: http://www.botsvsbrowsers.com/listings.asp?search=android%2F4.

aarontam:
Hi @wangii, we are going to close this PR in favor of https://github.com/enyojs/enyo/pull/1354. We have cherry-picked your original commit and made a few tweaks to fix merge conflicts and to combine the patterns into a single pattern. Thanks for your contribution!

PR #1016, opened 2015-01-29
ENYO-909: Only process CSS as LESS for resolution independent builds.

aarontam:
### Issue
We had made a change via https://github.com/enyojs/enyo/commit/b4d0f6 to process CSS files as LESS to support resolution-independence, but had failed to make this conditional on the resolution-independence option being set. As a result, standard builds would also try to process CSS files as LESS. This was problematic when there were errors in the CSS files (as a side-note, this effectively identified errors in the CSS files), and would cause the contents of that CSS file to not be included at all, while failing silently.

### Fix
We modify the conditional so that we only process CSS files if the resolution independence option has been set. Additionally, we throw an error if there is an issue processing our LESS files.

### DIFF
--- a/tools/minifier/minify.js
+++ b/tools/minifier/minify.js
@@ -102,11 +102,11 @@
 					w(" (Substituting CSS: " + sheet + ")");
 				}
 				var code = fs.readFileSync(sheet, "utf8");
-				if (isLess || isCss) {
+				if (isLess || (opt.ri && isCss)) {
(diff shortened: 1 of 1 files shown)


aarontam:
Closing this in favor of https://github.com/enyojs/enyo/pull/1017/ as the offending app will be updated in a timely fashion, and we want any errors, moving forward, to be fatal and thrown.

PR #1017, opened 2015-01-29
ENYO-910: Throw errors when using the LESS parser during deploy.

aarontam:
### Issue
We want any errors upon parsing CSS files to be fatal and thrown, to identify any CSS issues regardless of target system.

### Fix
Instead of logging the errors, we now throw them.


aarontam:
Closing per new build system.

PR #1018, opened 2015-01-30
ENYO-923: Stop marquee when window throw blur event

KunmyonChoi:
### Issue:
On app change state between BG and FG, div which is under cursor is not getting onleave event.
While marquee is flow on hover, it is not stop because it is not getting onleave event when app  goes to BG.

### Fix:
- Add blur on windowEvents array to handle on dispatcher feature. (Core)
- Remember last hovered marqueeSupport control on global variable. (Marquee)
- Call onleave event handler of last hovered marqueeSupport control when app is getting onblur event. (Marquee)

### DIFF
--- a/source/dom/dispatcher.js
+++ b/source/dom/dispatcher.js
@@ -42,7 +42,7 @@
 		*
 		* @private
 		*/
-		windowEvents: ["resize", "load", "unload", "message", "hashchange", "popstate", "focus", "blur"],
+		windowEvents: ["resize", "load", "unload", "message", "hashchange", "popstate"],
 
 		/**
 		* Feature plugins (aka filters)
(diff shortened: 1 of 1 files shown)


source/dom/dispatcher.js
```diff
@@ -54,8 +54,20 @@
 		/**
 		* @private
 		*/
+		platformSpecific: function() {
+			var d = enyo.dispatcher,
+				webosWindowEvents = ["focus", "blur"];
+			if (enyo.platform.webos >= 4) {
+				d.windowEvents.concat(webosWindowEvents);
```
clinuz: Array.prototype.concat method returns the new array and does not alter the arrays you're concatenating, so, this line is doing work but you aren't saving the returned array?

source/dom/dispatcher.js
```diff
@@ -54,8 +54,20 @@
 		/**
 		* @private
 		*/
+		platformSpecific: function() {
```
graynorton: The focus and blur events on the window object aren't actually specific to webOS, right? Is there any reason why we're trying to scope this change to webOS?

KunmyonChoi: I requested to get an event for app status change between FG and BG.
Browser team said I can use focus and blur event. I know that is comes when
I click input field also. But, webOs doesnt send focus event from window
object when focus on input field. So, the suggestion was to use focus blur
event comes from window object.

source/dom/dispatcher.js
KunmyonChoi: What a shame! Thanks for pointing out!

source/dom/dispatcher.js
KunmyonChoi: Or other way is we get window.focus/blur event here. But, not handling on spotlight when it is running on other than webOS platform. @graynorton how do you think?

KunmyonChoi:
Remove platform specific condition.
Instead, process only window event on spotlight and marquee.

PR #1021, opened 2015-02-03
ENYO-943: Support resolution-independence for CSS files in the browser.

aarontam:
### Issue
With the recent resolution-independence features, there was a lack of support for converting CSS of various sorts (non-LESS, essentially) for resolution-independence when loading in the browser.

### Fix
We modify `enyo.machine.sheet` to pass both CSS and LESS files found via `package.js` manifests to the LESS object for parsing. Additionally, we create our own "fork" of LESS with the `ri` suffix, which will allow for browser-based resolution-independence support for CSS files included via `link` tags, and also CSS found in  `style` blocks. Please see this commit for the actual diff (fairly minimal) between the current LESS distribution and the changes required for resolution-independence support:https://github.com/enyojs/enyo/commit/d33b356eb38cc4e922d8c7c3b3eff058e30a940a.

### Notes
I tried to find a path toward modularizing the requisite changes to the actual LESS source to parse CSS in the browser, but was unable to come up with anything clean, and worked under the assumption that we would own the version of the included LESS module and any references to it. Suggestions are more than welcome for this.


source/boot/boot.js
```diff
@@ -7,19 +7,18 @@
...
-				}
+			var isCss = (inPath.slice(-4) == '.css');
+			var isLess = !isCss || (inPath.slice(-5) == '.less');
+			if (window.less && (isLess || isCss)) {
+				// If client-side less is loaded, insert the less stylesheet
+				type = 'text/less';
+				rel = 'stylesheet/less';
+			} else if (isLess) {
+				// Otherwise, we expect a css file of the same name to exist
+				inPath = inPath.slice(0, inPath.length-4) + 'css';
```
Djspaceg: Should this be either `3` or `.css`?

aarontam: It seems to work as-is for files with the `.less` extension, which is only what the conditional will allow here. I actually didn't modify this line, despite what Github is telling you, it was just moved. :)

dmikey: We may want to add a note to the developer guide that CSS being parsed needs to adhere to less standards, terminating semi colons comes to mind (does this silently fail during build process?)

aarontam: Yep that's a good reminder for developers. Any LESS errors should be fatal during the build process after https://github.com/enyojs/enyo/pull/1017 is integrated.

Djspaceg: We should also consider turning on debugging when run from the debug.html file. So when errors DO happen, they're apparent. I wrote a stylesheet for an app that makes the debug output readable when used with a moonstone app which we can bake into enyo for consistency.

PR #1024, opened 2015-02-04
Update enyo.Select so it is bindable, and syncs

appsbychris:
Been using a lot of enyo.Select's in an app, i noticed the selected
property and value can get out of sync. Also, you could not bind to
selected. With these changes you can now bind to both selected and
value, and they stay in sync.

Also add properties for size and multiple

Enyo-DCO-1.1-Signed-off-by: Chris Van Hooser <chrisvanhooser@gmail.com>

### DIFF
--- a/source/ui/Select.js
+++ b/source/ui/Select.js
@@ -62,7 +62,25 @@
 			* @default null
 			* @public
 			*/
-			value: null
+			value: null,
+
+			/**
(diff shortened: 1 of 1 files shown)


ryanjduffy:
@appsbychris I pulled your changes into another PR with a few changes. Please take a look and let me know if you have any concerns. Appreciate the help!

PR #1025, opened 2015-02-04
ENYO-871: Imported resolution independence code from moonstone and added...

Djspaceg:
... some initialization and externalization features so any visual library can establish various screen resolutions their interface is targeting.

* Added `enyo.ri.defineScreenTypes`.
* Privatized the `screenTypes` array, and standardized the internal var names.

### DIFF
--- a/source/ui/resolution.js
+++ b/source/ui/resolution.js
@@ -189,7 +189,7 @@
 		init: function () {
 			_screenType = this.getScreenType();
 			this.updateScreenTypeOnBody();
-			enyo.dom.unitToPixelFactors.rem = scope.enyo.getUnitToPixelFactors();
+			enyo.dom.unitToPixelFactors.rem = this.getUnitToPixelFactors();
 			_screenTypeObject = getScreenTypeObject();
 			_riRatio = this.getRiRatio();
 		}
--- a/source/ui/resolution.js
+++ b/source/ui/resolution.js
@@ -1,0 +1,194 @@
+(function(enyo, scope) {
+
+	var _baseScreenType = 'standard',
+		_riRatio,
+		_screenType,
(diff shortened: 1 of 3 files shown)


source/ui/resolution.js
```diff
@@ -0,0 +1,194 @@
...
+		* This will need to be re-run any time the screen size changes, so all the values can be
+		* re-cached.
+		*
+		* @public
+		*/
+		// Later we can wire this up to a screen resize event so it doesn't need to be called manually.
+		init: function () {
+			_screenType = this.getScreenType();
+			this.updateScreenTypeOnBody();
+			enyo.dom.unitToPixelFactors.rem = scope.enyo.getUnitToPixelFactors();
```
aarontam: Should this be `enyo.ri.getUnitToPixelFactors()`?

source/ui/resolution.js
```diff
@@ -0,0 +1,194 @@
...
+			return 1;
+		},
+
+		/**
+		* @private
+		*/
+		getUnitToPixelFactors: function (type) {
+			type = type || _screenType;
+			if (type) {
+				_riRatio = getScreenTypeObject(type).pxPerRem;
```
aarontam: Is this line intentional - seems to be missing a divisor, potentially. If I call `enyo.ri.scale(1)` more than once, the result of the first call is `1` and subsequent calls are `24`, in FHD.

source/ui/resolution.js
Djspaceg: I caught this too late and it was corrected in 958. It should have been fixed in this branch though.

aarontam: Ah ha! Maybe I should have started with reviewing 958 instead - was trying to do them discretely to make the review more tractable. :smiley:

source/ui/resolution.js
Djspaceg: This does seem to be a general problem that didn't reveal itself earlier. I'll look into it.

PR #1031, opened 2015-02-09
Enyo 966 ChildSize changing does not affect on controlsPerPage.

DavidUM:
Issue
-------
In Notification app, dynamical changing of a control's height and refresh() call make blank.

Cause
--------
Updated childSize does not affect on controlsPerPage. When childSize is increased, controlsPerPage should be reduced because page size was not changed.
However, we still have same controlsPerPage and it reduce number of pages.
So, target page index could be reduced and it affect on wrong page adjustment.

Fix
----
Use static page size and update controlsPerPage when childSize is changed.

### DIFF
--- a/source/ui/data/VerticalDelegate.js
+++ b/source/ui/data/VerticalDelegate.js
@@ -48,6 +48,8 @@
 		* After size of viewport is decided, get pageSize using pageSizeMutiplier.
 		* We should update this.pageSize when we update viewport or pageSizeMutiplier.
 		*
+		* @param {enyo.DataList} list - The [list]{@link enyo.DataList} to perform this action on.
+		* @return {Number} this.pageSize - height or width of each page.
 		* @private
 		*/
 		calculatePageSize: function(list) {
@@ -55,6 +57,7 @@
 				multi = list.pageSizeMultiplier || this.pageSizeMultiplier;
 			// using height/width of the available viewport times our multiplier value
 			this.pageSize = fn.call(this, list) * multi;
+			return this.pageSize;
 		},
 		/**
 		* @private
--- a/source/ui/data/VerticalDelegate.js
+++ b/source/ui/data/VerticalDelegate.js
@@ -122,8 +122,10 @@
 		*/
 		refresh: function (list) {
 			if (!list.hasReset) { return this.reset(list); }
+			this._dirtyBit = true;
 			this.assignPageIndices(list);
 			this.generate(list);
+			this._dirtyBit = false;
 		},
 
 		/**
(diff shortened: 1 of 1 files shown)


source/ui/data/VerticalDelegate.js
```diff
@@ -45,9 +45,25 @@
...
+				multi = list.pageSizeMultiplier || this.pageSizeMultiplier;
+			// using height/width of the available viewport times our multiplier value
+			this.pageSize = fn.call(this, list) * multi;
+			return this.pageSize;
+		},
+		/**
 		* @private
 		*/
 		generate: function (list) {
+			this.pageSize = this.pageSize | this.calculatePageSize(list);
```
aarontam: Hi @DavidUM, did you mean to use `||` here instead of the bitwise operator?

source/ui/data/VerticalDelegate.js
```diff
@@ -236,12 +253,10 @@
 		* @private
 		*/
 		calculateControlsPerPage: function (list) {
-			var fn              = this[list.psizeProp],
-				multi           = list.pageSizeMultiplier || this.pageSizeMultiplier,
+			var pageSize        = this.pageSize | this.calculatePageSize(list),
```
aarontam: Similar to the above, did you mean to use `||` here instead?

source/ui/data/VerticalDelegate.js
```diff
@@ -191,7 +207,8 @@
 			metrics.height = this.pageHeight(list, page);
 			metrics.width  = this.pageWidth(list, page);
 			// update the childSize value now that we have measurements
-			this.childSize(list);
+			var oldChildSize = list.childSize;
+			list.isChildSizeUpdated = (oldChildSize == this.childSize(list)) ? false : true;
```
aarontam: Would it be better to move this, and related calculations, to `refresh`, so that we don't perform these calculations unnecessarily upon each page generation? From the app side, it seems that the developer is calling `refresh`, necessarily, when they know they have updated the size of the children.

DavidUM: Yes, good advise. But childSize can be updated after metrics is updated.
So, we can update childSize from end of generatePage which metric update is occurred.
Let me find another way to reduce unnecessary childSize() call.

aarontam: Ah ha, that makes sense - I think this will have to remain calculated here, I might have misread the diff. :)

source/ui/data/VerticalDelegate.js
```diff
@@ -236,12 +253,10 @@
 		* @private
 		*/
 		calculateControlsPerPage: function (list) {
-			var fn              = this[list.psizeProp],
-				multi           = list.pageSizeMultiplier || this.pageSizeMultiplier,
+			var pageSize        = this.pageSize || this.calculatePageSize(list),
```
aarontam: Will the page size always be updated as necessary here? It seems like the cached `this.pageSize` value will continue to be used even when we are in situations where we want it to be recalculated (i.e. viewport is resized).

PR #1050, opened 2015-02-27
ENYO-1098: Adding simulated flag on generated tap event

KunmyonChoi:
### Issue:
When (0, 0) is tapped on drawers, drawers do not open/close.
We have a code to filter out simulated click event in Spotlight onClick handler.
And, it is checking clientX/clientY == 0 to know is that a simulated event.
That is eventually filter out normal click event also.

### Fix:
Add a flag to indicate the event is generated from click event.
This can be used in Spotlight to exactly distinguish simulated event.

### DIFF
--- a/source/dom/dispatcher.js
+++ b/source/dom/dispatcher.js
@@ -288,7 +288,6 @@
 					// but note the tap event will fire first
 					var cp = enyo.clone(e);
 					cp.type = "tap";
-					cp.simulated = true;
 					cp.preventDefault = enyo.nop;
 					enyo.dispatch(cp);
 				}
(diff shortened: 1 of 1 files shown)


source/dom/dispatcher.js
```diff
@@ -283,11 +283,12 @@
 	enyo.dispatcher.features.push(
 		function (e) {
 			if ("click" === e.type) {
-				if (e.clientX === 0 && e.clientY === 0) {
+				if (e.clientX === 0 && e.clientY === 0 && e.detail === 0) {
```
aarontam: I'm wondering if this could be further simplified so that we only check `e.detail === 0` here as `e.clientX` and `e.clientY` are always 0 when `e.detail` is 0 as far as I can tell, or even change the conditional to `!e.detail` to fully support IE8 (which doesn't implement this property on `MouseEvent`). But pushed this commit for now so we can discuss.

PR #1051, opened 2015-02-27
ENYO-1013: Process image src whenever sizing changes, even if not yet generated

aarontam:
### Issue
This seems to have been broken for a long time, but setting the `sizing` property of `moon.Image` does not properly set the background-image as there is an ordering issue where the setting of the background-image is gated by the value of `sizing`, but in the current ordering of the bindings, the value of `sizing` would be set after the value of `src` at create time.

### Fix
We no longer guard the processing of the image `src` property and always process this when the value of `sizing` changes.

### DIFF
--- a/source/ui/Image.js
+++ b/source/ui/Image.js
@@ -204,8 +204,8 @@
 			if (this.sizing) {
 				this.addClass(this.sizing);
 			}
+			this.srcChanged();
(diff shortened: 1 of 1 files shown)


aarontam:
Closing in favor of https://github.com/enyojs/enyo/pull/1135.

PR #1052, opened 2015-03-03
Fixes ENYO-988 and ENYO-1032

ryanjduffy:
## Issue
enyo.Select.selected isn't a bindable property (as it should be) and setting the value of selected at create time is not respected.

## Fix
Some refactoring was needed to ensure the property changes were being applied in the right order and in the correct way (e.g. by using `set()` instead of modifying the member directly).

## Notes
This PR supercedes #1024. Thanks @appsbychris for much of the work to resolve this!

### DIFF
--- a/source/ui/Select.js
+++ b/source/ui/Select.js
@@ -62,7 +62,25 @@
 			* @default null
 			* @public
 			*/
-			value: null
+			value: null,
+
+			/**
+			* The size of the select box in rows.
+			* 
+			* @type {Number}
+			* @default 1
+			* @public
+			*/
+			size: 1,
+
+			/**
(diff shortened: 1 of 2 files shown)


source/ui/Select.js
```diff
@@ -62,7 +62,34 @@
...
+			value: null,
+
+			/**
+			* The size of the select box in rows.
+			* 
+			* @type {Number}
+			* @default 1
+			* @public
+			*/
+			size: 1,
```
aarontam: Should this be `0` to match the default specified in the spec: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select#Attributes.

appsbychris:
Awesome! thanks enyo-team!

PR #1056, opened 2015-03-04
ENYO-510-Auto show a placeholder in image grid as images load

rakesh25:
Issue:

There was no placeholder in image grid, and image gets displayed once loaded. In between grid remains empty.

Fix:

Using multiple background images through css, we can set the placeholder, a static image, which will be seen until the image gets loaded and will be replaced.

Enyo-DCO-1.1-Signed-off-by: Rakesh kumar rakesh25.kumar@lge.com

### DIFF
--- a/source/ui/Image.js
+++ b/source/ui/Image.js
@@ -231,10 +231,10 @@
 			this.tag = this.sizing ? 'div' : 'img';
 			this.addRemoveClass('sized', !!this.sizing);
 			if (was) {
-				this.removeClass('enyo-'+was);
+				this.removeClass(was);
 			}
 			if (this.sizing) {
-				this.addClass('enyo-'+this.sizing);
+				this.addClass(this.sizing);
 			}
 			if (this.generated) {
 				this.srcChanged();
--- a/source/ui/Image.js
+++ b/source/ui/Image.js
@@ -157,6 +157,10 @@
 			draggable: 'false'
 		},
 
+		handlers: {
+			onload: 'handleLoad'
+		},
+
 		/**
 		* @method
 		* @private
@@ -181,14 +185,17 @@
 		srcChanged: function () {
 			var src = enyo.ri.selectSrc(this.src);
 			if (this.sizing) {
-				var placeholder =  enyo.path.rewrite(this.placeholder);
-				var multiple= "url('"+ enyo.path.rewrite(src) +"'), url('"+ (placeholder ? placeholder : enyo.Image.placeholder) + "');";
-				this.applyStyle('background-image',  multiple);
+				var placeholder = this.placeholder ? ', url(\'' + enyo.path.rewrite(this.placeholder) + '\')' : '';
+				var url = 'url(\'' + enyo.path.rewrite(src) + '\')' + placeholder + ';';
+				this.applyStyle('background-image', url);
 			} else {
 				if (!src) {
 					// allow us to clear the src property
 					this.setAttribute('src', '');
 				} else {
+					if (this.placeholder) {
+						this.applyStyle('background-image', 'url(\'' + enyo.path.rewrite(this.placeholder) + '\')');
+					}
 					this.setAttribute('src', enyo.path.rewrite(src));
 				}
 			}
@@ -225,6 +232,15 @@
 		positionChanged: function () {
 			if (this.sizing) {
 				this.applyStyle('background-position', this.position);
+			}
+		},
+
+		/**
+		* @private
+		*/
+		handleLoad: function () {
+			if (!this.sizing && this.placeholder) {
+				this.applyStyle('background-image', null);
 			}
 		},
 
--- a/source/ui/Image.js
+++ b/source/ui/Image.js
@@ -187,8 +187,8 @@
 			var src = enyo.ri.selectSrc(this.src);
 			if (this.sizing) {
 				var placeholder = this.placeholder ? 'url(\'' + enyo.path.rewrite(this.placeholder) + '\')' : '';
-				var url= (src? 'url(\'' + enyo.path.rewrite(src) + '\'),' : '') + placeholder;
-				this.applyStyle('background-image', url);
+				var url= (src ? ( placeholder ? 'url(\'' + enyo.path.rewrite(src) + '\'),' : 'url(\'' + enyo.path.rewrite(src) + '\')') : '');
+				this.applyStyle('background-image', url+placeholder);
 			} else {
 				if (!src) {
 					// allow us to clear the src property
@@ -199,6 +199,15 @@
 					}
 					this.setAttribute('src', enyo.path.rewrite(src));
 				}
+			}
+		},
+
+		/**
+		* @private
+		*/
+		placeholderChanged: function(){
+			if (this.placeholder) {
+				this.applyStyle('background-image', 'url(\'' + enyo.path.rewrite(this.placeholder) + '\')');
 			}
 		},
 
--- a/source/ui/Image.js
+++ b/source/ui/Image.js
@@ -126,7 +126,14 @@
 			* @default 'center'
 			* @public
 			*/
-			position: 'center'
+			position: 'center',
+
+			/**
+			* @type {String}
+			* @default ''
+			* @public
+			*/
+			placeholder: ''
 		},
 
 		/**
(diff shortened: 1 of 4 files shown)


source/ui/Image.js
```diff
@@ -173,14 +180,15 @@
 		*/
 		srcChanged: function () {
 			var src = enyo.ri.selectSrc(this.src);
+			var multiple= "url('"+ enyo.path.rewrite(src) +"'), url('"+ this.placeholder + "');";
```
aarontam: I think we should guard against a falsy value of `placeholder` here.

source/ui/Image.js
```diff
@@ -173,14 +180,15 @@
...
 			if (this.sizing) {
-				this.applyStyle('background-image', src ? 'url(' + enyo.path.rewrite(src) + ')' : 'none');
+				this.applyStyle('background-image', src ? multiple : 'none');
 			} else {
 				if (!src) {
 					// allow us to clear the src property
 					this.setAttribute('src', '');
 				} else {
-					this.setAttribute('src', enyo.path.rewrite(src));
+					this.applyStyle('background-image', multiple );
```
aarontam: I think we have to carefully consider how we're approaching this. With the change on this line, we would always be setting the background-image on this element, and the `src` attribute would never be set. Previously, we were only setting the background-image when `sizing` had been specified, so we have to determine how the value of `placeholder` fits into this. For example, if we specify `placeholder` but not `sizing`, do we still set background-image? If so, what kind of sizing would we default to? We could use a sensible default in this scenario, i.e. `cover`, so that the entire background is covered. We just want to make sure we don't alter the existing behavior and allow the `src` property to still be set. We should also make it clear that specifying a value for `placeholder` relies on using background-image.

source/ui/Image.js
aarontam: In order to properly support placeholder image assets, we will also need to rewrite the asset url (for the case where an SVG data URI is not specified).

rakesh25: I will do the required changes to check for falsy value of `placeholder`. And if no placeholder defined, then default placeholder to be used. 
About proper support of placeholder image assets, Does someone from Enyo SVL will look into `rewrite` aspect or I have to come up with something to support.

source/ui/Image.js
rakesh25: I agree with the approach mentioned. I will do the requisite changes.

source/ui/Image.js
aarontam: Thanks! I was thinking that to properly support image paths set with the `placeholder` property, we would just need to wrap the value in a call to `enyo.path.rewrite`, similar to how we were originally handling the `src` property, to account for $paths and such.

source/ui/Image.js
```diff
@@ -174,7 +181,9 @@
 		srcChanged: function () {
 			var src = enyo.ri.selectSrc(this.src);
 			if (this.sizing) {
-				this.applyStyle('background-image', src ? 'url(' + enyo.path.rewrite(src) + ')' : 'none');
+				var placeholder =  enyo.path.rewrite(this.placeholder);
+				var multiple= "url('"+ enyo.path.rewrite(src) +"'), url('"+ (placeholder ? placeholder : enyo.Image.placeholder) + "');";
```
aarontam: I don't know if we always want a placeholder image displayed - this should be something developers opt into at the {{enyo.Image}} level, so leaving the value as the default empty string shouldn't display a placeholder.

source/ui/Image.js
```diff
@@ -174,12 +186,17 @@
 		srcChanged: function () {
 			var src = enyo.ri.selectSrc(this.src);
 			if (this.sizing) {
-				this.applyStyle('background-image', src ? 'url(' + enyo.path.rewrite(src) + ')' : 'none');
+				var placeholder = this.placeholder ? ', url(\'' + enyo.path.rewrite(this.placeholder) + '\')' : '';
+				var url = 'url(\'' + enyo.path.rewrite(src) + '\')' + placeholder + ';';
```
aarontam: I think we need to guard against a falsy value such as an empty string here, otherwise we'll set the `background-image` to `url('')`, which results in an error as we are trying to set the background-image to the current page's URL:

```
Resource interpreted as Image but transferred with MIME type text/html
```

source/ui/Image.js
```diff
@@ -174,12 +186,17 @@
 		srcChanged: function () {
 			var src = enyo.ri.selectSrc(this.src);
 			if (this.sizing) {
-				this.applyStyle('background-image', src ? 'url(' + enyo.path.rewrite(src) + ')' : 'none');
+				var placeholder = this.placeholder ? 'url(\'' + enyo.path.rewrite(this.placeholder) + '\')' : '';
+				var url= (src ? ( placeholder ? 'url(\'' + enyo.path.rewrite(src) + '\'),' : 'url(\'' + enyo.path.rewrite(src) + '\')') : '');
```
ryanjduffy: Not keen on the double ternary operator here and it causing you to duplicate the rewrite call.

source/ui/Image.js
```diff
@@ -174,12 +186,17 @@
...
+				this.applyStyle('background-image', url+placeholder);
 			} else {
 				if (!src) {
 					// allow us to clear the src property
 					this.setAttribute('src', '');
 				} else {
+					if (this.placeholder) {
+						this.applyStyle('background-image', 'url(\'' + enyo.path.rewrite(this.placeholder) + '\')');
+					}
```
ryanjduffy: This block is duplicating `placeholderChanged'. While we could replace it with a call to that method, I think it'd be better to add a call in `create` before `srcChanged` which would give the same effect. More importantly, by adding the call in `create`, we also allow for setting `placeholder` and not `src`. With the current code, nothing would be displayed (if both were set at create-time) because this block would not be hit.

source/ui/Image.js
```diff
@@ -174,12 +186,17 @@
 		srcChanged: function () {
 			var src = enyo.ri.selectSrc(this.src);
 			if (this.sizing) {
-				this.applyStyle('background-image', src ? 'url(' + enyo.path.rewrite(src) + ')' : 'none');
```
Djspaceg: Removing 'none' removes the ability for one to "unset" the background image. This would mean that the property would be removed and whatever the property inherited would take prescience, so it would be impossible to take away the image, and only restore it to what some CSS had applied earlier.

source/ui/Image.js
Djspaceg: This does seem excessive, running `enyo.path.rewrite` so many times. This should be simplified, and only run when necessary.

source/ui/Image.js
```diff
@@ -188,6 +205,15 @@
 		/**
 		* @private
 		*/
+		placeholderChanged: function(){
+			if (this.placeholder) {
```
Djspaceg: There's no way to unset the placeholder's style here.

rakesh25: In case of *img* tag, when there is no source set/exists, i.e. *!this.src* returns *true*, then only, I feel, unset/changed placeholder's value makes sense. If image source already exists, changing placeholder will remove the set source on img. 
Talking about *div* case, approach used considers multiple background image url's. If at later point of time, placeholder needs to be unset, then it will be done only when no source value exists.
Let me know your thoughts, if using some other approach we can achieve unsetting placeholder, no matter if src exists or not.

source/ui/Image.js
rakesh25: Right..Will do the requisite changes!

source/ui/Image.js
rakesh25: `enyo.path.rewrite` will be called only ONCE, based on placeholder value existence.

source/ui/Image.js
```diff
@@ -199,10 +231,10 @@
 			this.tag = this.sizing ? 'div' : 'img';
 			this.addRemoveClass('sized', !!this.sizing);
 			if (was) {
-				this.removeClass(was);
+				this.removeClass('enyo-'+was);
 			}
 			if (this.sizing) {
-				this.addClass(this.sizing);
+				this.addClass('enyo-'+this.sizing);
```
Djspaceg: My mistake, and apologies. These should not have the "enyo-" prefix. I was remembering incorrectly.

rakesh25: Right!...Made it back to what it was.

ryanjduffy:
Closing master PRs

PR #1062, opened 2015-03-05
Fixes ENYO-1140

ryanjduffy:
## Issue
The less.js injection script assumes it will be located in <app-root>/enyo/tools and therefore hard codes the path to the actual LESS script relative to <app-root>. If enyo is in a shared location (or were moved into lib), the path would be incorrect.

## Fix
Determine path to LESS based on src of LESS loader script tag


aarontam:
@ryanjduffy this is great - we should be able to update the Moonstone samples to take advantage of this change, once it's been integrated.

ryanjduffy:
Good call. Forgot about the ugly paths they all have. Did you open a ticket for that already?

aarontam:
Nope not yet - I had created a branch with an updated set of samples, but ran into this issue you've resolved, in the course of the resolution-independence work. I'll create a PR. :+1:

aarontam:
@ryanjduffy just remembered we had created a separate `less-ri.js` file so that developers could opt-out from using the resolution-independent version of LESS (i.e. https://github.com/enyojs/bootplate-moonstone/blob/master/debug.html#L16). Can we also make the same updates to this script as well?

lishammel:
After discussion with Ryan, closing out as the jira associated with this was fixed another way.

PR #1074, opened 2015-03-17
ENYO-1220: Support /lib/enyo location for Enyo path and minifier

JayCanuck:
Enyo-DCO-1.1-Signed-off-by: Jason Robitaille <jason.robitaille@lge.com>


clinuz:
Closing this as it is against master and we will not be accepting it for future releases.

PR #1075, opened 2015-03-18
ENYO-1200 Update partially filled page when models are added.

DavidUM:
Issue
------
A page generated once, and it is not fully filled page.
And some models are added without re-generate that page.
We we call scrollToIndex over that page, generatePage() is called infinitely.

Cause
--------
Model adding does not update partially filled page's bounds

FIx
-----
The last page of list can be partially filled if models are not equally distributed to every page.
However if new models are added and then list has more pages than before,
previous last page must be fully filled.

Currently that last page does not update its metric until it is generated again.
So, list.bufferSize could miss calculate it.
We can prevent this edge case with this commit.

### DIFF
--- a/source/ui/data/VerticalDelegate.js
+++ b/source/ui/data/VerticalDelegate.js
@@ -413,14 +413,20 @@
 			if (props.index <= end ) this.refresh(list);
 			else {
 				// we should confirm that the page which new models are added is need to update list.metrics
-				var targetPageIndex = this.pageForIndex(list, props.index),
-					page = list.metrics.pages[targetPageIndex],
+				var lastPageIndex = this.pageForIndex(list, props.index),
+					// the last page before model added
+					lastPage = list.metrics.pages[lastPageIndex],
 					sp = list.psizeProp,
+					// current page count after models added
 					pc = this.pageCount(list),
-					pageSize = this.defaultPageSize(list);
+					pageSize;
 
-				if (targetPageIndex < pc && page[sp] < pageSize) {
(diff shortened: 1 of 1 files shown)


source/ui/data/VerticalDelegate.js
```diff
@@ -401,13 +401,28 @@
...
 			if (props.index <= end ) this.refresh(list);
 			else {
+				// we should confirm that the page which new models are added is need to update list.metrics
+				var targetPageIndex = this.pageForIndex(list, props.index),
+					page = list.metrics.pages[targetPageIndex],
+					sp = list.psizeProp,
+					pc = this.pageCount(list),
+					pageSize = this.defaultPageSize(list);
+
+				if (targetPageIndex < pc && page[sp] < pageSize) {
```
clinuz: According to the bug description, wouldn't this only need to happen if the page was the last page? If it wasn't the last page, then we know it is already full. In that case, we wouldn't need to update this value or call `this.pageCount` or `this.defaultPageSize` needlessly. We also wouldn't need to check `page[sp] < pageSize` and could safely just assign the value (would fix the issue in problem case and be a nop in regular case).

DavidUM: Exactly as what you said. 
{{targetPageIndex}} is an index of the last page before model added. I may change its name.

This logic will be executed only once when bug description is happened.
If page count is not increased after model added, or the last page is already fully filled
then we do not need to update its metric.

Let me add one more commit for better readability. Thanks!

PR #1067, opened 2015-03-10
ENYO-1122: Add "accessibilityLabel" to enyo control

jaewonjang:
The "accessibilityLabel" property covers ARIA attribute "aria-label".
If "accessibilityLabel" is set, add "aria-label" and make the control
tabbable by setting a tabindex=0.
Screen reader reads "aria-label" when the control is focused.

https://jira2.lgsvl.com/browse/ENYO-1122
Enyo-DCO-1.1-Signed-off-by: Jaewon Jang <jaewon98.jang@lgepartner.com>

PR #1078, opened 2015-03-20
ENYO-1188: Accessibility: Apply "aria-label" to all enyo control which h...

jaewonjang:
...ave a content

When content of control is set or chagned, aria-label should be also set
and changed. So I added the code of setting aria-label in contentChanged
function.
In addition, it needs the tabindex attribute to be focusable because
screen reader can read aria-label when dom is focused. And if dom has
tabindex attribute default outline is shown when it focused, so remove
the outline in case dom has tabindex attribute.

https://jira2.lgsvl.com/browse/ENYO-1188

### DIFF
--- a/source/dom/dom.css
+++ b/source/dom/dom.css
@@ -139,4 +139,6 @@ body .enyo-fullscreen {
 *[tabindex] {
 	/* remove outline in case dom has tabindex attribute */
 	outline: none;
+	/* remove tap highlight in case dom has tabindex attribute */
+	-webkit-tap-highlight-color: rgba(0, 0, 0, 0);
 }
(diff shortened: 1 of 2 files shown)


jaewonjang:
This PR has dependency on https://github.com/enyojs/enyo/pull/1067

source/dom/dom.css
```diff
@@ -134,3 +134,9 @@ body .enyo-fullscreen {
 	height: 100% !important;
 	box-sizing: border-box !important;
 }
+
+/* Accessibility CSS */
+*[tabindex] {
+	/* remove outline in case dom has tabindex attribute */
+	outline: none;
```
seunghoh: Don't we need to put -webkit-tap-highlight to disable browser's default highlight when tap?

jaewonjang: Tap highlight is only shown on the browser based on mobile webkit, so I think it should be added on garnet library.

seunghoh:
Looks good to me

jaewonjang:

PR #1084, opened 2015-03-24
Add enyo Scrim that was skiped/missed by LG

webOS101:
Enyo-DCO-1.1 Signed-off-by: Signed-off-by: johnmcconnell@yahoo.com


clinuz:
Closing because this will have no meaning moving into 2.6 and beyond.

PR #1085, opened 2015-03-25
ENYO-1124: Add "accessibilityAlert" to enyo control

jaewonjang:
The accessibilityAlert property is to cover alert role of aria
attribute.
Basically screen reader reads content or accessibilityLabel out when the
control is focused, but there are some cases that it should be read
something regardless focus. For example, popup, notifications, page
description and so on. If accessibilityAlert is set true, screen reader
will automatically reads content or accessibilityLabe when the control
is rendered or visible.
Refer
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_alert_role.

https://jira2.lgsvl.com/browse/ENYO-1124

### DIFF
--- a/source/dom/Control.js
+++ b/source/dom/Control.js
@@ -215,6 +215,21 @@
 		accessibilityLabel : '',
 
 		/**
+		* AccessibilityAlert is for alert message or page description.
+		* If accessibilityAlert is true, screen reader will automatically reads
+		* content or accessibilityLabel regardless focus.
+		*
+		* Range: [`true`, `false`]
+		* - true: screen reader automatically reads label regardless focus.
+		* - false: screen reader reads label with focus.
+		*
+		* @type {Boolean}
+		* @default false
+		* @public
+		*/
+		accessibilityAlert : false,
+
+		/**
 		* @todo Find out how to document "handlers".
 		* @public
 		*/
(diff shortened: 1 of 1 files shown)


jaewonjang:
This PR has dependency on #1067

source/dom/Control.js
```diff
@@ -215,6 +215,21 @@
 		accessibilityLabel : '',
```
seunghoh: We don't have space after property name.
accessibilityLabel: '',

source/dom/Control.js
```diff
@@ -215,6 +215,21 @@
...
+		*
+		* Range: [`true`, `false`]
+		* - true: screen reader automatically reads label regardless focus.
+		* - false: screen reader reads label with focus.
+		*
+		* @type {Boolean}
+		* @default false
+		* @public
+		*/
+		accessibilityAlert : false,
```
seunghoh: We don't have space after property name.
accessibilityAlert: false

dmikey:
Perhaps a more generic 'role' property would be better, with a default value. There are more roles than just accessibilityAlert, for popups we'll most likely need to use alertdialog instead - when a user needs to take action on a popup.

jaewonjang:

PR #1095, opened 2015-04-01
ENYO-510 Enhancements

ryanjduffy:
There were a couple scenarios in which `placeholder` and `src` didn't play well together.

* If sizing and src were specified but placeholder wasn't, the generated CSS was invalid (it had a trailing comma after the url because it assumed a following url for the placeholder).
* If placeholder was specified but src wasn't, the placeholder was not rendered. This was a logic gap in srcChanged

Also, when sizing, src, and placeholder were all specified, the placeholder image wasn't removed because `handleLoad` isn't called when the image is loaded as a CSS `background-image`. To handle this, I set the `background-image` as the placeholder and I added a proxy `` to which the handlers are bound that loads the src. Once the src loads, the `background-image` is updated to be the src and the proxy is removed.


rakesh25:
PR https://github.com/enyojs/enyo/pull/1056
The first two scenarios you mentioned are handled now properly in PR mentioned above. Regarding  removing the background-image when the src image gets loaded, includes one more complexity of loading backgroung-image, whenever "src" is updated. That's why I didn't consider removing background-image in my PR. Moreover I think it could be a performance issue considering there will be separate image element creation for each grid item. Can you please share your views on this.
Thank you!

ryanjduffy:
There certainly is a performance impact to the additional node but I'd question how great it truly is. If we wanted to use this approach, we could further optimize it to not destroy the Image so when the tag is reused in a DataList, we wouldn't be unnecessarily creating and destroying them.

The simpler alternative is to document that when using `placeholder`, `src`, and `sizing` together, the placeholder image may show through depending on the `sizing` used and the image's characteristics.

Commented on your PR with a couple suggestions. I'd also recommend you cherry pick 81fe1bf from here to flesh out the Image Sample.

PR #1096, opened 2015-04-01
ENYO-948 (enyojs JIRA ENYO-4129)

clinuz:
# Issue

When inlining uri-encoded utf8 SVG data into `background-image`, it is possible to encounter nested IRIs (e.g. `url()`) that are improperly re-written by our current `minify.js` build script.

# Fix

Modify the regular expression to properly avoid recursive searches and skip cases we don't want to rewrite and only then rewrite what we need to.


clinuz:
This needs to be against 2.5-upkeep, not master.

PR #1100, opened 2015-04-03
ENYO-944: Adding flux support.

aarontam:
Cherry-picking the changes from the `enyo-flux-derekanderson` branch for integration into `ENYO-944`.

### DIFF
--- a/source/data/FluxStore.js
+++ b/source/data/FluxStore.js
@@ -1,0 +1,121 @@
+(function (enyo, scope) {
+
+	var kind = enyo.kind,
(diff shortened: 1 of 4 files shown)


source/data/FluxStore.js
```diff
@@ -0,0 +1,190 @@
...
+		/**
+		* @private
+		*/
+		kind: enyo.Object,
+
+
+		/**
+		* @private
+		*/
+		data: {},
```
aarontam: Should this be initialized in `create` instead, to prevent object reference sharing?

PR #1101, opened 2015-04-03
ENYO-1308: Placeholder background color

KunmyonChoi:
- Add a backgroundColor property into enyo.Image.
- Add a condition to skip unnecessary DOM access on create

### DIFF
--- a/samples/ImageSample.js
+++ b/samples/ImageSample.js
@@ -5,7 +5,11 @@ enyo.kind({
 		{content: "Image", classes: "section"},
 		{kind: "enyo.Image", src: "http://enyojs.com/img/enyo-logo.png", alt: "Enyo Logo"},
 		{content: "Background Image", classes: "section"},
-		{kind: "enyo.Image", style: "width:200px;height:75px;background-color:gray", sizing: "constrain", position: "left top", src: "http://enyojs.com/img/enyo-logo.png", alt: "Enyo Logo"},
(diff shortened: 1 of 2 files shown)


aarontam:
Hi @KunmyonChoi, we are implementing something along the same lines for ENYO-510 (see https://github.com/enyojs/enyo/pull/1095, https://github.com/enyojs/enyo/pull/1056, https://github.com/enyojs/layout/pull/123, and https://github.com/enyojs/moonstone/pull/2010). I wonder if we shouldn't try to consolidate this work with the other feature and/or at least take into account the techniques being used there?

KunmyonChoi:
Hi @aarontam , actually I tried also the onload technique for placeholder color. Because, I found that onload event only comes when I am using img tag but not comes from background-image. What I found is set src first then switch src to background-image on onload handler. I think @ryanjduffy 's technique is also do the same. I asked to GUI designer about the exact requirement to make the requirement simple. And she said that it doesn't need to be reset the background color after load. Which means image can be displayed with background color even when image has transparent background.

aarontam:
Ah ha, makes sense! I think another part of my question was considering the case when `placeholder` and `backgroundColor` are both set, and which gets precedence, and other potential interference cases.

lishammel:
@KunmyonChoi  - do we need this PR anymore? The Jira is closed as "will not fix". Thanks.

PR #1102, opened 2015-04-03
ENYO-1125: Add "accessibilityHint" to enyo control

jaewonjang:
AccessibilityHint is used for additional information of control. If
accessibilityHint is set and content exists, screen reader reads
accessibilityHint with content when control is focused.
For example there is a button and if button clicked, image or video will
be played. In this case blind needs to know what will happen after
clicking button.
AccessibilityHint can explain the result of action and describe
something need to more information.

https://jira2.lgsvl.com/browse/ENYO-1125

### DIFF
--- a/source/dom/Control.js
+++ b/source/dom/Control.js
@@ -212,7 +212,18 @@
 		* @default ''
 		* @public
 		*/
-		accessibilityLabel : '',
+		accessibilityLabel: '',
+
+		/**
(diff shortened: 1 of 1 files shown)


jaewonjang:
This PR has dependency on https://github.com/enyojs/enyo/pull/1078

jaewonjang:

PR #1108, opened 2015-04-13
ENYO-1227 : Tap event is not fired when scroll thumb is not disappeared in enyo.Scroller

kbs12e:
## Issue

Tap event is not fired when scroll thumb is not disappeared in Scroller. For details, 
even though Scroller can not scroll up(or down) because its touchoverscroll value is false, tap event is prevented. then when scroll job process is ended, scroll thumb is hidden and user's event is triggered. so the users think that tap event callback is not just worked.
## Cause

If user taps scroller during scrolling, ondown event stopped scroll and tap event is prevented. 
This is scroller's intended action. but even though Scroller can not scroll up(or down) because its touchoverscroll value is false, Scroller calls internally scroll processes. so, scroll thumb is shown and prevent tap event.
## Fix

The scroll callback function moves scroller and calculates its boundary. so, when boundary is scroll edge, just stop the scroll job because it does not needs to progress scroll job.

### DIFF
--- a/source/touch/Scroller.js
+++ b/source/touch/Scroller.js
@@ -233,7 +233,7 @@
 		* @default true
 		* @public
 		*/
-		touchOverscroll: false,
+		touchOverscroll: true,
 
 		/**
 		* If `true`, the [scroller]{@link enyo.Scroller} will not propagate `dragstart`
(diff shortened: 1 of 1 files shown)


lishammel:
Closing PR.

PR #1115, opened 2015-04-15
ENYO-1441 : modify resolution update routine for removing old resolution classes

sbcho:
## issue
moon.ri.init() function needs a line for removing ScreenTypeOnBody
moon.ri.getScreenTypeObject always return old screenTypeObject if there is no parameter

## fix
added oldScreenType variable for for checking and removing old resolution classes.
and removed caching routine in getScreenTypeObject function.
(this is a copy of https://github.com/enyojs/moonstone/pull/2074 for master branch)

Enyo-DCO-1.1-Signed-off-by: Sungbae Cho sb.cho@lge.com

### DIFF
--- a/source/dom/resolution.js
+++ b/source/dom/resolution.js
@@ -7,9 +7,9 @@
 		_screenTypeObject,
 		_oldScreenType;
 
-	var getScreenTypeObject = function (type, ignoreCache) {
+	var getScreenTypeObject = function (type) {
 		type = type || _screenType;
-		if (type == _screenType && !ignoreCache && _screenTypeObject) {
+		if (_screenTypeObject && _screenTypeObject.name == type) {
 			return _screenTypeObject;
 		}
 		return _screenTypes.filter(function (elem) {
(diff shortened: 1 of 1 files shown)


source/dom/resolution.js
```diff
@@ -4,13 +4,11 @@
...
 		_screenTypes = [ {name: 'standard', pxPerRem: 16, width: scope.innerWidth,  height: scope.innerHeight, aspectRatioName: 'standard'} ],	// Assign one sane value in case defineScreenTypes is never run.
-		_screenTypeObject;
+		_screenTypeObject,
+		_oldScreenType;
 
 	var getScreenTypeObject = function (type) {
 		type = type || _screenType;
-		if (type == _screenType && _screenTypeObject) {
-			return _screenTypeObject;
-		}
```
Djspaceg: Why was this optimization removed? (Reasoning) It prevents an array search if the request is the same as the current screenType.

sbcho: Oh there was 2 reason why I remove it.
- if there is no parameter ({{type}}), type variable will be assigned as _screeType value and {{type == _screenType}} will be always true. But we can't assure _screeTypeObject is updated sync with screeType object.(it can return old resolution _screeTypeObject)
- this function seems to don't called very frequently so I thought it can have an array type search.

If second reason looks not reasonable, I'll change it as another way

PR #1129, opened 2015-04-21
ENYO-1124: Accessibility: Add "accessibilityAlert" to enyo control

jaewonjang:
The accessibilityAlert property is to cover alert role of aria
attribute.
Basically screen reader reads content or accessibilityLabel out when the
control is focused, but there are some cases that it should be read
something regardless focus. For example, popup, notifications, page
description and so on. If accessibilityAlert is set true, screen reader
will automatically reads content or accessibilityLabel when the control
is rendered or visible.
Refer

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_alert_role.

https://jira2.lgsvl.com/browse/ENYO-1124
Enyo-DCO-1.1-Signed-off-by: Jaewon Jang jaewon98.jang@lgepartner.com

### DIFF
--- a/source/dom/accessibility.js
+++ b/source/dom/accessibility.js
@@ -26,6 +26,21 @@
 			accessibilityHint: '',
 
 			/**
+			* AccessibilityAlert is for alert message or page description.
+			* If accessibilityAlert is true, screen reader will automatically reads
+			* content or accessibilityLabel regardless focus.
+			*
+			* Range: [`true`, `false`]
+			* - true: screen reader automatically reads label regardless focus.
+			* - false: screen reader reads label with focus.
+			*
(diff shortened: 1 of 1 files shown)


source/dom/accessibility.js
```diff
@@ -26,6 +26,21 @@
 			accessibilityHint: '',
 
 			/**
+			* AccessibilityAlert is for alert message or page description.
+			* If accessibilityAlert is true, screen reader will automatically reads
+			* content or accessibilityLabel regardless focus.
+			*
```
seunghoh: Please also add remark that this will replace "role" to "alert"

PR #1134, opened 2015-04-28
ENYO-2013 fixing store subscribe

dmikey:
Store should subscribe to it's own notification chain, if it has an update method defined.


aarontam:
Hi @dmikey it looks like this change has effectively already gone into `2.6.0-dev`: https://github.com/enyojs/enyo/blob/2.6.0-dev/lib/FluxStore.js#L114. Can we close this out?

dmikey:
Yup!

PR #1137, opened 2015-04-29
ENYO-931: removed unnecessary trailing semi-colon

rajyavardhanp:
DCO-1.1-Signed-Off-By: Rajyavardhan P <rajyavardhan.p@lge.com>


lib/Control/Control.js
```diff
@@ -1663,7 +1663,7 @@ Control.concat = function (ctor, props, instance) {
 
 	if (props.style) {
 		if (instance) {
-			str = (proto.style ? (proto.style + ';') : '') + (props.style + ';');
+			str = (proto.style ? proto.style : '') + props.style;
```
aarontam: Will this open us up to the possibility of style concatenation without a semicolon between the style properties of the parent kind and the child kind (i.e. if the developer leaves off the trailing semicolon in the parent kind's `style` property)? The appearance of double semicolons was not an issue before because we would run the concatenated style string through the `normalizeCssStyleString` method: https://github.com/enyojs/enyo/blob/2.4.0/source/dom/Control.js#L1106.

rajyavardhanp: Hi @aarontam , I tested the solution with different combinations of parent kind's styles, its working fine in all the cases. I guess by design of 'control concat' methods, it doesn't create any issues. However  it is causing problems in `overrideComponets` block, I don't know exact reason, some analysis is required to identify the problem. But in either case your solution works, as other things will be taken care by `normalizeCssStyleString`

aarontam: Is the issue with `overrideComponents` directly related to your change here? If not, then I think this should be good to go. You're right that by running `normalizeCssStyleString` at the end of this conditional, we don't open ourselves up to the possibility of not having missing semicolons.

PR #1150, opened 2015-05-04
Fixes ENYO-1553

ryanjduffy:
## Issue
Options couldn't be selected using their `selected` property at create time. It was overridden by the logic in `Select.rendered()`.

## Fix
Support selecting Option with selected on either Option or Select.
Added tests for Option and Select

### DIFF
--- a/lib/Select.js
+++ b/lib/Select.js
@@ -20,7 +20,7 @@ var
 *	{content: 'Ascending', value: 'a'}
 * ]}
 * 
-* selectChanged: function(inSender, inEvent) {
+* selectChanged: function (inSender, inEvent) {
 *	var s = inSender.getValue();
 *	if (s == 'd') {
 *		this.sortListDescending();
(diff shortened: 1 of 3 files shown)


lib/Select.js
```diff
@@ -125,14 +125,19 @@ module.exports = kind(
...
 				this.setAttribute('onchange', dispatcher.bubbler);
 			}
-			this.selectedChanged();
-			this.valueChanged();
+			// This makes Select.selected a higher priority than Option.selected but ensures that
+			// the former works at create time
+			if (this.selected !== 0) {
+				this.selectedChanged();
+			} else {
+				this.updateSelectedFromNode();
```
aarontam: Hi @ryanjduffy, when the conditional is not met, we have a case where `this.selected` is `0` and the value is not subsequently set, because `selectedChanged` is not run due to the `selectedIndex` property of the control also being `0`.

ryanjduffy: @aarontam Fixed!

PR #1154, opened 2015-05-06
ENYO-1405: Migrate Lite panel concept to New panel design (Initial commi...

KunmyonChoi:
...t)

- Apply index queue
- Add extension point for BreadcrumbSupport

### DIFF
--- a/lib/LightPanels/LightPanels.js
+++ b/lib/LightPanels/LightPanels.js
@@ -484,7 +484,7 @@ module.exports = kind(
 	resetView: function (view) {
 		// reset position
 		var trans = {};
-		trans['translate' + this._axis] = '0%';
+		trans['translate' + this._axis] = 100 * this._direction + '%';
 		dom.transform(view, trans);
 	},
 
@@ -666,7 +666,7 @@ module.exports = kind(
 	applyTransitions: function (nextPanel, direct) {
 		// apply the transition for the next panel
 		var nextTransition = {};
-		nextTransition['translate' + this._axis] = -100 * this._direction + '%';
+		nextTransition['translate' + this._axis] = '0%';
 		dom.transform(nextPanel, nextTransition);
 		if (this._currentPanel) { // apply the transition for the current panel
 			this.shiftPanel(this._currentPanel, this._indexDirection);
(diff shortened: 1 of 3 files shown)


lib/LightPanels/LightPanels.js
```diff
@@ -166,15 +171,33 @@ module.exports = kind(
...
 			this.indexChanged();
 		};
 	}),
 
+	/**
+	* We set initial index as 0 when there is static components block.
+	* @private
+	*/
+	determineInitialIndex: function() {
+		if (this.index = -1 && this.getPanels().length > 0) {
```
aarontam: I think you probably meant `this.index == -1` :)

lib/LightPanels/LightPanels.js
```diff
@@ -306,7 +355,7 @@ module.exports = kind(
 			this.index = newIndex;
 			this.setupTransitions(currentIndex, false);
 		} else {
-			this.set('index', newIndex);
+			this.set('index', newIndex, true);
```
aarontam: Is it necessary to force a change (now that we've changed the default value of `index` back to `-1`)?

lib/LightPanels/LightPanels.js
```diff
@@ -195,15 +219,40 @@ module.exports = kind(
...
 	/**
 	* @private
 	*/
 	indexChanged: function (was) {
+		// Blocking wrong index change
+		if (this.index < 0 || this.index > this.getPanels().length - 1) {
+			this.index = was;
+			return;
+		}
+		this.rememberQueueIndex(was);
```
aarontam: Is it possible to only make this call if we have mixed-in the breadcrumb mixin?

lib/LightPanels/LightPanels.js
```diff
@@ -539,6 +595,26 @@ module.exports = kind(
 	*/
 
 	/**
+	* @private
+	*/
+	adjustDuration: function(diff) {
```
aarontam: It seems like this method could be moved to the mixin.

lib/LightPanels/LightPanels.js
```diff
@@ -539,6 +595,26 @@ module.exports = kind(
...
+		if (this._duration === undefined) {
+			this._duration = this.duration;
+		}
+		this.duration = this.duration / Math.abs(diff);
+	},
+	
+	/**
+	* @private
+	*/
+	restoreDuration: function() {
```
aarontam: Can we also move this method into the mixin?

PR #1160, opened 2015-05-12
ENYO-1609: setContent(0) on item is not changing content when it has …

KunmyonChoi:
…initial value of empty string

Issue:
We are comparing content with previous content with != on setContent.
It is considering '' and 0 as the same.

Fix:
We are compare content with type checking by using '!==' instead of '!='.


KunmyonChoi:
Close PR per original issue is closed.

PR #1164, opened 2015-05-13
ENYO-1629: Add support for naming of managers. General clean-up.

aarontam:
### Issue
We need support for the naming of managers, so that they can be referenced in the future. Also, there is some technical debt where we have some logic for determining framerate in `BackgroundTaskManager`, which should be moved to `SystemMonitor`.

### Fix
We've added a `managerName` property in `TaskManagerSupport`, and have updated `BackgroundTaskManager` to maintain a hash of names and customers. Additionally, we have moved the framerate checking logic from `BackgroundTaskManager` to `SystemMonitor`.

### DIFF
--- a/lib/BackgroundTaskManager.js
+++ b/lib/BackgroundTaskManager.js
@@ -6,8 +6,7 @@ require('enyo');
 */
 
 var
-	kind = require('./kind'),
-	utils = require('./utils');
+	kind = require('./kind');
 
 var
 	CoreObject = require('./CoreObject'),
@@ -58,37 +57,12 @@ module.exports = kind.singleton(
 	*/
 	constructor: kind.inherit( function (sup) {
 		return function () {
-			var c, p, f = 0, d = 1000 / this.fpsThreshold;
-
 			sup.apply(this, arguments);
 
-			this.cb = this.bindSafely(function() {
-				if (this.customers.length) {
-					if (!SystemMonitor.active) {
-						SystemMonitor.trigger();
-					}
+			SystemMonitor.on('idle', this.notifyIdle, this);
 
-					if (!c) {
-						c = utils.perfNow();
-					} else {
-						p = c;
-						c = utils.perfNow();
-						f = ((c - p) < d) ? f + 1 : 0;
-					}
-					if (f == this.frameThreshold && SystemMonitor.idle()) {
-						this.run();
-						c = p = f = 0;
-					} else {
-						// reset fps check if threshold is met but system is not user-idle
-						if (f == this.frameThreshold) {
-							c = p = f = 0;
-						}
-						this.trigger();
-					}
-
-				} else if (SystemMonitor.active) {
-					SystemMonitor.stop();
-				}
+			this.cb = this.bindSafely( function () {
+				if (!SystemMonitor.get('active')) SystemMonitor.start();
 			});
 		};
 	}),
@@ -96,8 +70,8 @@ module.exports = kind.singleton(
 	/**
 	* @private
 	*/
-	trigger: function() {
-		Loop.request(this.cb);
+	trigger: function () {
+		if (this.hasActiveTasks()) Loop.request(this.cb);
 	},
 
 	/**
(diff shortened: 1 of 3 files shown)


ryanjduffy:
I don't follow why we need to defer activating customers and preserve removing them? What are we gaining here?

aarontam:
The thinking behind this was to not waste cycles in continuously running through a queue of TaskManagers that had no tasks; we would only add a TaskManager to the run queue once it had at least one task. And related to that, if we wanted to access a TaskManager by name that had no tasks in its queue, preserving it in the BTM would allow for access from anywhere. The first item might be a pre-optimization, while the last item seems like it could be handled differently with the modular system, now. Thoughts?

ryanjduffy:
I don't have a strong objection to the current design but here's an alternative to consider. It does appear that leaving customers without tasks could unnecessarily defer their work because `BackgroundTaskManager.run()` plucks the first customer and tries to run its task. If it doesn't have any, that cycle is wasted. Instead, if we had a `hasTasks()` method to query if there was work to do (this would return false when `this.tasks.length === 0` and maybe when `this.paused`), we could iterate over the customers until we found one with work and then do it. Just want to make sure we need to add (possibly confusing) methods to the public API.

Related, rather that shifting and pushing the customer off of and onto the array, you may want to maintain an index of the current customer. That would save you the array manipulation but would cause you to have to manage the index when customers are removed before the index.

lib/SystemMonitor.js
```diff
@@ -55,25 +72,61 @@ module.exports = kind.singleton(
...
+					p = c;
+					c = utils.perfNow();
+					f = ((c - p) < d) ? f + 1 : 0;
+				}
+
+				if (f == this.frameThreshold && this.idle()) {
+					this.active = false;
+					this.emit('idle');
+					if (this.listeners('idle').length === 0) {
+						var idx = dispatcher.features.indexOf(this.bindSafely(this.checkEvent));
```
ryanjduffy: I don't think this will work as you expect. Each call to `bindSafely` returns a new function so you'll never find a previous one with the result of another call.

aarontam: Yep you're right! Ran into this when storing the popStateHandler in {{moon.History}} as well - let me make an update, thanks!

lib/BackgroundTaskManager.js
```diff
@@ -46,39 +45,24 @@ module.exports = kind.singleton(
...
-						this.trigger();
-					}
-
-				} else if (SystemMonitor.active) {
-					SystemMonitor.stop();
-				}
+			SystemMonitor.on('idle', this.notifyIdle, this);
+
+			this.cb = this.bindSafely( function () {
+				if (!SystemMonitor.get('active')) SystemMonitor.start();
```
ryanjduffy: Doesn't appear you need to bind the function since you're not using any local context. Assuming that's just cruft from the previous code. Also, since `SystemMonitor.start()` has a check already for `active`, that's probably not necessary here. I think that would make more sense anyway as it's probably not the caller's responsibility to know if SystemMonitor is currently active before calling `start()`.

lib/TaskManagerSupport.js
```diff
@@ -155,7 +181,7 @@ module.exports = {
 			this.task && this.task();
 
 			if (this.tasks.length === 0) { // remove ourselves if we no longer have tasks
-				BackgroundTaskManager.remove(this);
+				BackgroundTaskManager.remove(this.managerName, true);
```
ryanjduffy: One 1 arg now

lib/TaskManagerSupport.js
```diff
@@ -59,7 +69,9 @@ module.exports = {
 		return function () {
 			sup.apply(this, arguments);
 			this.tasks = new PriorityQueue();
-			this._namedTasks = {};
+			this.namedTasks = {};
+			this.managerName = this.managerName || this.name;
+			BackgroundTaskManager.add(this, this.managerName, true);
```
ryanjduffy: Only 1 arg now

PR #1169, opened 2015-05-19
Enyo 1452 Use accurate childSize to prevent blank page

DavidUM:
Issue
------
DataList show blank page when actual childSize is less than 100px

Cause
-------
Without fixedChildSize, DataList could not know what is actual childSize. 
So, childSize() uses assumed childSize like 100px
A controlsPerPage is decided based on assumed childSize and PageSize is calculated with childSize and controlsPerPage.  
Here is clue of blank issue.

To guarantee proper page adjusting behavior, pageSize should be larger than viewport size.
Because of that, we have pageSizeMultiplier.
However, if actual childSize is smaller than assumed childSize, pageSize could be smaller than viewport size.

Fix
----
Rendering a single child before we generate full pages is enough condition to have accurate childSize.

### DIFF
--- a/source/ui/data/VerticalDelegate.js
+++ b/source/ui/data/VerticalDelegate.js
@@ -192,7 +192,14 @@
 				perPage = this.controlsPerPage(list);
 
 			// the first index for this generated page
-			page.start  = perPage * index;
+			// if childSize is just updated, we do not need to re-render
+			// the first child of this page.
+			if (list.childSizeNeedsUpdate) { 
+				page.start = perPage * index + 1;
+				list.childSizeNeedsUpdate = false;
+			} else {
+				page.start  =  perPage * index;	
+			}			
 			// the last index for this generated page
 			page.end    = Math.min((data.length - 1), (page.start + perPage) - 1);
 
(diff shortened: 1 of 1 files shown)


source/ui/data/VerticalDelegate.js
```diff
@@ -214,18 +233,33 @@
...
+						// render the first page and get accurate childSize
+						if (list.controlsPerPage) {
+							page.start = page.end = list.controlsPerPage * page.index;	
+						} else {
+							page.start = page.end = 0;
+						}
+						
+						this.updatePage(list, page);
+						childSize = page.getBounds()[sizeProp];
+						list.childSizeNeedsUpdate = (list.childSize != childSize)
```
yeram: ';' is missing at the end of this line.

source/ui/data/VerticalDelegate.js
```diff
@@ -214,18 +233,33 @@
 		* @private
 		*/
 		childSize: function (list) {
-			if (!list.fixedChildSize) {
-				var pageIndex = list.$.page1.index,
+			if (!list.fixedChildSize || !list.childSize) {
+				var page = list.$.page1,
 					sizeProp  = list.psizeProp,
 					n         = list.$.page1.node || list.$.page1.hasNode(),
```
yeram: We can use 'page' instead of 'list.$.page1'

source/ui/data/VerticalDelegate.js
```diff
@@ -186,6 +169,42 @@
...
+				// the metrics for the entire list
+				metrics = list.metrics,
+				// controls per page
+				perPage = this.controlsPerPage(list);
+
+			// the first index for this generated page
+			// if childSize is just updated, we do not need to re-render
+			// the first child of this page.
+			if (list.childSizeNeedsUpdate) { 
+				page.start = perPage * index + 1;
```
yeram: Question, how about (page.start == page.end) case?

source/ui/data/VerticalDelegate.js
DavidUM: For example, if there is only a single model in collection?
Currently, {{updatePage}} request from generatePage is ignored inside of updatePage logic.
However, it is not proper code. Let me fix it.

Thank you for your review. :+1:

DavidUM:
I close this PR but if we still have dataList in 2.7, we'd better consider this again.

PR #1173, opened 2015-05-21
Fixes ENYO-1677

ryanjduffy:
## Issue
In the new module system, there's no obvious need to have singletons as their purpose was to make a globally available instance. Instead, the kind can be instanced and exposed from the module. Now, the singletons are being exported into the global namespace unnecessarily potentially allowing usage without importing the required module (which is fragile).

# Fix
Mark the function deprecated and remove usage

### DIFF
--- a/lib/AccessibilitySupport/AccessibilitySupport.js
+++ b/lib/AccessibilitySupport/AccessibilitySupport.js
@@ -35,7 +35,7 @@ module.exports = {
 
 	/**
 	* AccessibilityAlert is for alert message or page description.
-	* If accessibilityAlert is true, aria role will be set to "alert" and 
+	* If accessibilityAlert is true, aria role will be set to "alert" and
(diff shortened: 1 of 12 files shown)


lib/NewAnimator.js
```diff
@@ -1,16 +1,14 @@
...
 var
-	kind = require('./kind'),
 	utils = require('./utils');
 
 var
 	CoreObject = require('./CoreObject'),
 	Loop = require('./Loop');
 
-module.exports = kind.singleton({
+module.exports = CoreObject({
```
aarontam: Should this be `new CoreObject` instead?

ryanjduffy: Ya. Discovered that I need to rework this PR anyway as you can't extend existing methods with this approach ... you just overwrite them.

aarontam: Ahh that's unfortunate! :-1:

PR #1174, opened 2015-05-22
ENYO-1624: DataList Record value Hides after the ExpandableInput Collapses

anishlg:
### Issue 
The moon Datalist recordCount is not getting displayed in some cases.

### Cause 1:
When the Datalist is not scrolled it will generate only 2 pages. The recordCount value became more than 389 means the previous index (props.index) is more than 38 which result  the lastPageIndex will more than 1 (pagesPerControl  = 19) in the VerticalDelegate modelsAdded () function. This will result wrong page index.
### Solution: 
Use end value to calculate the lastPageIndex instead of previous index (props.index)  if the DataList having only 2 pages ( not scrolled).

### Cause 2:
The previous index (props.index) and end value is based on count will always give incremented lastPageIndex in the boundary condition. This is boundary value issue
### Solution:
Reduce the end value and pros.index value by 1 to get the proper index.

DCO-1.1-Signed-Off-By: Anish TD anish.td@lge.com

### DIFF
--- a/lib/VerticalDelegate.js
+++ b/lib/VerticalDelegate.js
@@ -409,7 +409,8 @@ module.exports = {
 
 		var cpp = this.controlsPerPage(list),
 			end = Math.max(list.$.page1.start, list.$.page2.start) + cpp - 1,
-			idx = props.index - 1;
+			idx = props.index,
+			count = Object.getOwnPropertyNames(list.metrics.pages).length * cpp - 1;
 
 		// note that this will refresh the following scenarios
 		// 1. if the dataset was spliced in above the current indices and the last index added was
@@ -437,7 +438,7 @@ module.exports = {
 		if (idx <= end ) this.refresh(list);
 		else {
 			// we should confirm that the page which new models are added is need to update list.metrics
-				var lastPageIndex = Object.getOwnPropertyNames(list.metrics.pages).length == 2 ?   this.pageForIndex(list, end) : this.pageForIndex(list, idx),
+				var lastPageIndex = this.pageForIndex(list, count),
 				// the last page before model added
 				lastPage = list.metrics.pages[lastPageIndex],
 				sp = list.psizeProp,
--- a/lib/VerticalDelegate.js
+++ b/lib/VerticalDelegate.js
@@ -408,7 +408,8 @@ module.exports = {
 		if (!list.hasReset) return this.reset(list);
 
 		var cpp = this.controlsPerPage(list),
-			end = Math.max(list.$.page1.start, list.$.page2.start) + cpp;
+			end = Math.max(list.$.page1.start, list.$.page2.start) + cpp - 1,
+			idx = props.index - 1;
 
 		// note that this will refresh the following scenarios
 		// 1. if the dataset was spliced in above the current indices and the last index added was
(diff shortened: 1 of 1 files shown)


lib/VerticalDelegate.js
```diff
@@ -408,7 +408,8 @@ module.exports = {
 		if (!list.hasReset) return this.reset(list);
 
 		var cpp = this.controlsPerPage(list),
-			end = Math.max(list.$.page1.start, list.$.page2.start) + cpp;
+			end = Math.max(list.$.page1.start, list.$.page2.start) + cpp - 1,
```
DavidUM: Good point. 
{{end}} means last index of pages, so we should subtract 1.
Thank you @anishlg

lib/VerticalDelegate.js
```diff
@@ -408,7 +408,8 @@ module.exports = {
...
+			idx = props.index - 1;
```
DavidUM: But in this line, I have doubt for `-1`
`props.index` means starting index of added models.
If we had 38 models and then add 100 models, `props.index` is 38.

lib/VerticalDelegate.js
```diff
@@ -433,10 +434,10 @@ module.exports = {
 
 		// if we need to refresh, do it now and ensure that we're properly setup to scroll
 		// if we were adding to a partially filled page
-		if (props.index <= end ) this.refresh(list);
+		if (idx <= end ) this.refresh(list);
```
DavidUM: If starting index of added models is greater than `end` which is last index of pages, we do not need to generate pages.
But, if it is equal or less than `end`, current pages should be updated.
I think you don't need to subtract 1.

lib/VerticalDelegate.js
```diff
@@ -433,10 +434,10 @@ module.exports = {
...
 		else {
 			// we should confirm that the page which new models are added is need to update list.metrics
-			var lastPageIndex = this.pageForIndex(list, props.index),
+				var lastPageIndex = Object.getOwnPropertyNames(list.metrics.pages).length == 2 ?   this.pageForIndex(list, end) : this.pageForIndex(list, idx),
```
DavidUM: What we really need is `accurate page metric` of not generated page.
You assume that this issue could be reproducible only we generated 2 pages.
But If there are some not generate page, we can meet this issue again.

lib/VerticalDelegate.js
anishlg: the pros.index means previous model count. so if the previous model count is 38 it will have 2 pages with 19 models each and ideally should return 1 as last page index. but if we pass 38 which divided by 19 will give 2. this is I meant by boundary condition. if pros.index is 0-18 -> page 0, 19-37-> page 1 38-56 -> page 3 and so on.

lib/VerticalDelegate.js
anishlg: ya I got your point. I will try to get the accurate page metric than the hard coded length checking

lib/VerticalDelegate.js
DavidUM: Yes. props.index could be previous model count in this scenario. 
But basically, second argument of `modelsAdded()` means newly added model's properties.
Not for existed models. 

Anyway `props.index` indicates the first index which new models are added.
That the reason why we compare it with `end`.
If `props.index` is in current pages, we should update them.
It is line 436.

lib/VerticalDelegate.js
DavidUM: Thank you for your works!

anishlg:
Updated with review comments.
1. retain the props.index value intact.
2. calculate the accurate page metrics

graynorton:
Closing in favor of #1219.

PR #1175, opened 2015-05-22
ENYO-1685: Add try/catch on notify to guard unexpected error.

KunmyonChoi:
Issue:
If paths of computed property have an observer. And the observer function meet error for some reason.
Then the computed value will never flushed again. Because '_computedRecursion' will be never get to 0.

Fix:
We add try/catch to catch error on super call of notify then recover _computedRecursion to 0.
So that it can safely flushComputed even after error.


clinuz:
For a code path executed as often as this it should be noted that using try/catch will keep the compiler from ever optimizing the code. This should definitely be fixed by some other means.

KunmyonChoi:
Close PR per there is still no reported issue.

PR #1176, opened 2015-05-26
ENYO-1648: Allow for combining and resetting of global holdpulse config.

aarontam:
### Issue
Previously the default holdPulseConfig would be overwritten by a call to `configureHoldPulse`, allowing no way of resetting to the default. Additionally, because the configuration was set by assignment, rather than mixed-in, the full configuration had to be specified when calling `configureHoldPulse`, rather than just the properties to override.

### Fix
We now mixin the passed-in config when calling `configureHoldPulse`. Additionally, a `resetHoldPulseConfig` method has been added. Also, updated the JSHint configuration to expose `JSON` as a global var, to prevent Travis failures.

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -4,7 +4,8 @@
     "require": false,
     "exports": true,
     "module": true,
-    "module.exports": true
+    "module.exports": true,
+    "JSON": true
   },
   "es3": true,
   "evil": false,
--- a/lib/gesture/drag.js
+++ b/lib/gesture/drag.js
@@ -134,9 +134,22 @@ var drag = module.exports =
 	*
 	* @public
 	*/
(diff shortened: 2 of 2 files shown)


.jshintrc
```diff
@@ -4,7 +4,8 @@
     "require": false,
     "exports": true,
     "module": true,
-    "module.exports": true
+    "module.exports": true,
+    "JSON": true
```
KunmyonChoi: Hi @aarontam , do you know why we need this change here?

aarontam: Hi @KunmyonChoi, so it appears that using the JSON global *should* have caused an error previously. We either started using it somewhere, or something changed with JSHint. I didn't have a chance to track this one down completely.

PR #1188, opened 2015-06-02
ENYO-1689: Accessibility: migration accessibilityDisabled to enyo 2.6

jaewonjang:
The accessibilityDisabled is to be disable voice readout for control.
In some requirements, it should not read anything when control is
focused, so we provide accessibilityDisabled property to cover that.
If accessibilityDisabled is true, aria attributes are removed for the
control.
Note that this is not working on HTML form elements which can get focus
without tabindex.

https://jira2.lgsvl.com/browse/ENYO-1689

### DIFF
--- a/lib/AccessibilitySupport/AccessibilitySupport.js
+++ b/lib/AccessibilitySupport/AccessibilitySupport.js
@@ -72,8 +72,6 @@ module.exports = {
 	* Note that this is not working on HTML form elements which can get focus without tabindex.
 	*
 	* Range: [`true`, `false`]
-	* - true: screen reader doesn't read control label.
-	* - false: screen reader reads control label.
 	*
 	* @type {Boolean}
 	* @default false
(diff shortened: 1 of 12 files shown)


lib/AccessibilitySupport/AccessibilitySupport.js
```diff
@@ -67,13 +67,30 @@ module.exports = {
 	accessibilityLive: false,
 
 	/**
+	* AccessibilityDisabled prevents VoiceReadout.
+	* If accessibilityDisabled is true, screen reader doesn't read any label for the control.
+	* Note that this is not working on HTML form elements which can get focus without tabindex.
+	*
+	* Range: [`true`, `false`]
```
ryanjduffy: This comment isn't necessary

lib/AccessibilitySupport/AccessibilitySupport.js
```diff
@@ -100,6 +117,10 @@ module.exports = {
 		var prefix = this.accessibilityLabel || this.content || null;
 		var label;
 
+		if (this.accessibilityDisabled) {
```
ryanjduffy: This is a bit a personal preference but I'd rather see these checks as one line without the braces. No need for the extra lines of code for a kick out.

```
if (this.accessibilityDisabled) return;
```

ryanjduffy:
Closing in favor of #1204

PR #1193, opened 2015-06-05
Router: ignore self-induced hashchanges

johannjacobsohn:
When changing the location hash in the Router via this.location(newHash), the hashchange listeners are triggered. This is not always what we want since we've dealt with the state change already.


source/kernel/Router.js
```diff
@@ -221,6 +231,11 @@
 		*/
 		_history: null,
 
+		/*
+		* @private
+		*/
+		_selfChanged: false,
```
aarontam: Can probably remove this property as it's not used.

aarontam:
Hi @johannjacobsohn, thanks for your contribution! I'm curious about your implementation that leads to the logic always being run both before and in response to a hash change event. Would you be able to modify the logic and/or add some guards?

Regarding the fix itself, I think the overall intent seems reasonable, but the behavior seems like it could be fairly conditional in many use cases, and it might make more sense to enable this feature via a parameter when calling `location`, rather than setting a property on the `Router` instance itself. Otherwise, if we wanted this behavior to be more of a fundamental feature, we might consider some refactoring at the `hashchange` listener level.

lishammel:
After additional internal discussion, we are deciding not to merge in this fix.  @johannjacobsohn - please let us know if you would like to discuss further.

PR #1194, opened 2015-06-08
ENYO-1725: enyo.Video: setPlaybackRate should be called ahead of play in a certain case

suhlee:
## Issue
If user selects REW/FF when content is paused, setPlaybackRate should be called ahead of play.

## Fix
added flag, "isNeedPlay", to make node play after setPlaybackRate  .

Enyo-DCO-1.1-Signed-off-by: Suhyung Lee (suhyung2.lee@lge.com)

### DIFF
--- a/src/Video.js
+++ b/src/Video.js
@@ -499,7 +499,8 @@ module.exports = kind(
 	* @public
 	*/
 	fastForward: function () {
-		var node = this.hasNode();
+		var node = this.hasNode(),
+			isNeedPlay = false;
 
 		if (!node) {
 			return;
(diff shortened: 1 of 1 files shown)


lib/Video.js
```diff
@@ -550,6 +551,7 @@ module.exports = kind(
 
 		this.setPlaybackRate(this.selectPlaybackRate(this._speedIndex));
 
+		isNeedPlay && node.play();
```
aarontam: Not a huge fan of these types of statements (but open for debate). I think it'd be clearer and not many characters to do something like this:
```
if (isNeedPlay) node.play();
```

PR #1197, opened 2015-06-09
Refactor logic to update image src

ryanjduffy:
* Combine all `src` and `background-image` updates into a single method for consistent application which caught a couple edge cases where those were not updated correctly.
* Proxy `src` with `_src` so we don't overwrite the user-provided value with our error handler
* Added some docs while I was there

### DIFF
--- a/lib/Image/Image.js
+++ b/lib/Image/Image.js
@@ -41,6 +41,11 @@ var
 * [onerror]{@link enyo.Image#onerror} [events]{@glossary event}. Image dragging is suppressed by
 * default, so as not to interfere with touch interfaces.
 *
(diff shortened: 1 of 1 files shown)


lib/Image/Image.js
```diff
@@ -172,12 +183,22 @@ module.exports = kind(
...
 	handlers: {
 		onload: 'handleLoad',
 		onerror: 'handleError'
 	},
 
 	/**
+	* @private
+	*/
+	observers: [
+		{method: 'updateSource', path: ['_src', 'placeholder']}
```
rajyavardhanp: placeholder is to be changed only if there is no source, I guess. If source is already loaded and we want to change just the place holder calling 'updateSource' doesn't seem a good thing in UI perspective.

PR #1205, opened 2015-06-10
ENYO-1795: Enable configuration of resolution-independence plugin in browser.

aarontam:
### Issue
There did not exist a mechanism to specify configuration options for the LESS resolution-independence plugin when the script is loaded in the browser.

### Fix
As long as `less.js` is being loaded synchronously, we can parse the query string of its `src` path by examining the last script tag in the document. This *should* work for the case of `bootplate-moonstone` (and various related `bootplate` implementations) as we have control over the default implementation and how the script is being loaded. Also, cleaned up the option parsing a bit.

### DIFF
--- a/tools/less.js
+++ b/tools/less.js
@@ -1,24 +1,40 @@
 // Client-side less.js configurator / loader
 (function() {
+	// parse script for options
+	var scripts = document.getElementsByTagName('script'),
+		index = scripts.length - 1,
+		riParams = scripts[index].getAttribute('ri-params'),
+		opts;
+
+	if (riParams) {
+		opts = {};
+		riParams.replace(
+			new RegExp('([^,=&]+)(=([^,]*))?', 'g'),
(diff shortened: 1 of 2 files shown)


ryanjduffy:
I think an attribute on the script tag would be a better and more standard option here than query string parameters. wdyt?

aarontam:
Are you thinking of a single attribute that has all of the parameters shoehorned in there, or a single attribute for each parameter? I'd be amenable to the first option from an implementation perspective, as we don't need to worry about knowing the set of applicable attributes, though I wonder if the ergonomics would be awkward (though probably no more awkward than query string parameters).

ryanjduffy:
1st option would be fine, imo

aarontam:
Sounds good - updated!

aarontam:
@ryanjduffy, squashed all the commits so make sure to do a hard pull, thanks!

PR #1207, opened 2015-06-15
ENYO-1764: Remove 'aria-label' on controls witch added as default wit…

jaewonjang:
…h content value

As per blink engine turned out to support read content without adding
'aria-label', we will remove aria-label with same value as content. Enyo
control will have only tabindex=0 for supporting accessibility.
aria-label will be used only when we need to read something else.

https://jira2.lgsvl.com/browse/ENYO-1764

### DIFF
--- a/test/tests/AccessibilitySupport.js
+++ b/test/tests/AccessibilitySupport.js
@@ -31,13 +31,13 @@ describe('AccessibilitySupport', function () {
 				TestControl = null;
 			});
 
-			it ('should equal content', function () {
+			it ('should equal null with only content', function () {
 
 				testControl.set('content', content);
 				testControl.set('accessibilityLabel', '');
 				testControl.set('accessibilityHint', '');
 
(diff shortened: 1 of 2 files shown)


seunghoh:
1st reviewed and looks good to me. 
I've discussed with Gray on this issue and decided that we will add only tabindex=0 if we need to read only content. aria-label will be added only when app developer calls 'accessibilityLabel' or 'accessibilityHint' or both of them

test/tests/AccessibilitySupport.js
```diff
@@ -37,7 +37,7 @@ describe('AccessibilitySupport', function () {
 				testControl.set('accessibilityLabel', '');
 				testControl.set('accessibilityHint', '');
 
-				expect(testControl.getAttribute('aria-label')).to.equal(content);
+				expect(testControl.getAttribute('aria-label')).to.equal(null);
```
ryanjduffy: Pls update the test label ('should equal content') to match the test code.

PR #1209, opened 2015-06-15
ENYO-1832: Fix typo that caused utils and dispatcher to be in the global scope.

aarontam:
### Issue
Both `utils` and `dispatcher` were in the global scope.

### Fix
There was a typo where we had inadvertently used a semi-colon instead of a comma in a group `var` statement. The semi-colon has now been replaced with the comma.
 

ryanjduffy:
SEMICOLONS!

PR #1210, opened 2015-06-16
Fixes ENYO-1840

ryanjduffy:
## Issue
`$L` and `updateLocale` live in enyo/hooks which isn't obvious to devs.

## Fix
Rename the module to enyo/i18n and add extensibility features so we're not just stomping functions from other modules/libraries.


ryanjduffy:
Ug ... wrong branch ...

PR #1211, opened 2015-06-16
Fixes ENYO-1840

ryanjduffy:
## Issue
`$L` and `updateLocale` live in enyo/hooks which isn't obvious to devs.

## Fix
Rename the module to enyo/i18n and add extensibility features so we're not just stomping functions from other modules/libraries.


lib/utils.js
```diff
@@ -19,31 +19,74 @@ if (!Function.prototype.bind) {
...
 		ret.prototype = new nop();
 		/*jshint +W055 */
-		
+
 		return ret;
 	};
 }
 
 /**
+* Returns a function that can be extended (via `.extend()`) similar to prototypal inheritance.
```
aarontam: :+1:

PR #1212, opened 2015-06-18
ENYO-1285: Marquee ellipse 

dmikey:
Issue.

Children are not notified of a parents reflow event, but are re-rendered, which causes their own reflow. 

Fix.

Notify children of reflow, so that they do not need to be re-rendered to take reflow action


dmikey:
No loner needed!

PR #1213, opened 2015-06-18
Enyo 1596 jaewonjang

jaewonjang:
In case webos devices we just call webOS.voicereadout.readAlert api
using TTS.
On none webos devices, we append a dom node which has alert role to read
alert message, after then remove it.

https://jira2.lgsvl.com/browse/ENYO-1596

### DIFF
--- a/lib/readAlert.js
+++ b/lib/readAlert.js
@@ -25,21 +25,21 @@ function readAlert (s) {
 	} else {
 		removeAlertChild();
 		appendAlertChild(s);
-		setInterval(removeAlertChild, 300);
+		setTimeout(removeAlertChild, 300);
 	}
 }
 
 function appendAlertChild (s) {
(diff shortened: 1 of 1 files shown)


lib/readAlert.js
```diff
@@ -0,0 +1,46 @@
...
+	var isWebos = window.webOS && window.webOS.voicereadout && window.webOS.platform &&
+					window.webOS.platform.watch ||
+					window.webOS.platform.tv ||
+					null;
+	if (isWebos) {
+		window.webOS.voicereadout.readAlert(s);
+	} else {
+		removeAlertChild();
+		appendAlertChild(s);
+		setInterval(removeAlertChild, 300);
```
seunghoh: Why setInerval call is needed? You should call clearInterval for clearing timer, or timer will accumulate whenever you call readAlert

PR #1216, opened 2015-06-24
ENYO-1880: Add animateTo function.

aarontam:
### Issue
We needed to initiate a panel index change to an existing panel, without a transition animation.

### Fix
We did some refactoring and made a standard panel index change a non-animated function, and added an explicit `animateTo` function for initiating an animated panel index transition.

### DIFF
--- a/lib/LightPanels/LightPanels.js
+++ b/lib/LightPanels/LightPanels.js
@@ -199,7 +199,7 @@ module.exports = kind(
 	* @private
 	*/
 	indexChanged: function (was) {
-		this.setupTransitions(was, false);
+		this.setupTransitions(was);
 	},
 
 	/**
--- a/lib/LightPanels/LightPanels.js
+++ b/lib/LightPanels/LightPanels.js
@@ -199,7 +199,7 @@ module.exports = kind(
 	* @private
 	*/
 	indexChanged: function (was) {
-		this.setupTransitions(was, this.shouldAnimate());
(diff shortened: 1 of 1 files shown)


lib/LightPanels/LightPanels.js
```diff
@@ -247,6 +256,21 @@ module.exports = kind(
...
 	/**
+	* Changes to the specific index, accepting a number of additional options.
+	*
+	* @param {Number} to - The index of the panel we wish to transition to.
+	* @param {module:enyo/LightPanels~UpdateIndexOptions} [opts] - Additional options to be used
+	*	when updating the panel index.
+	* @param {Number} [from] - If known, the index we are transitioning from.
+	* @public
+	*/
+	updateIndex: function (to, opts, from) {
```
aarontam: I wanted to consolidate the logic from `indexChanged` so it would be in one location, but resulted in slightly awkward ergonomics for `updateIndex`. That being said, only the internal `indexChanged` would realistically utilize the third parameter (`from`). Let me know if this is too awkward in general and we can move to use an `updateIndexDirect` method, which I was trying really hard to avoid. :smile:

ryanjduffy:
Probably worth adding some documentation indicating that changing the index will not animate and that the `animate` property applies to pushing new panels or navigating via `next()` and `previous()`.

aarontam:
@ryanjduffy Yep good call - I added some notes about the updated behavior.

PR #1218, opened 2015-06-25
DRD-10027 & ENYO-2003  Marquee is not starting for "Revoke all pairing authorizations" pop-up.

dmikey:
Issue.

So the font is slightly bigger, and that's the reason why we see an elipsis. The problem for the font being slightly bigger is caused by using -webkit-font-kerning: normal. This is ignored in blink. If you use font-kerning: normal it will work.

Fix.

Created a new normalizing plugin for Less.JS that will go through the rule sets, and create missing vendor prefixes or non-prefixed versions of the rules.


aarontam:
We don't want to include all those node modules, right? :smile:

dmikey:
I'm guessing not? I'm not sure how 2.6 works with this, I suspect they need to come along for the ride with 2.5 though?

webOS101:
Is that a node module that should be installed with `npm install`?  Is it in the package.json?  Maybe that is the way to include it, but I don't know.

aarontam:
Yeah I would imagine they would be dependencies (possibly of another dependency, here) that would be installed via `npm install`, and not need to be explicitly included.

dmikey:
we don't use npm install for any of the 2.5 build tools however.

PR #1222, opened 2015-07-01
ENYO-1981: Prevent pushing panels when transitioning. General clean-up.

aarontam:
### Issue
When panels are transitioning, panels can still be pushed, which causes undesired or unexpected behavior.

### Fix
We set a `transitioning` flag and guard against pushing panels when transitioning. We also do some general clean-up and removal of cruft.

### DIFF
--- a/lib/LightPanels/LightPanels.js
+++ b/lib/LightPanels/LightPanels.js
@@ -16,11 +16,12 @@ var
 	TaskManagerSupport = require('../TaskManagerSupport');
 
 var
-	LightPanel = require('./LightPanel');
+	LightPanel = require('./LightPanel'),
+	States = LightPanel.States;
 
 /**
 * @enum {Number}
-* @memberof module:moonstone/LightPanels~LightPanels
+* @memberof module:enyo/LightPanels~LightPanels
 * @public
 */
 var Direction = {
@@ -30,7 +31,7 @@ var Direction = {
 
 /**
 * @enum {Number}
-* @memberof module:moonstone/LightPanels~LightPanels
+* @memberof module:enyo/LightPanels~LightPanels
 * @public
 */
 var Orientation = {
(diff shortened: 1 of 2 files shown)


lib/LightPanels/LightPanels.js
```diff
@@ -524,29 +548,22 @@ module.exports = kind(
...
+			if (this.popQueue && this.popQueue.length) this.finalizePurge();
+			this.transitioning = false;
+		}
 
-			if (this._previousPanel) {
-				this._previousPanel.removeClass('transitioning');
-				this._previousPanel.deactivated && this._previousPanel.deactivated();
-			}
+		panel.removeClass('transitioning');
+		if (panel.postTransition) {
```
ryanjduffy: These checks aren't necessary since enyo/LightPanel contains a no-op implementation of the method.

lib/LightPanels/LightPanels.js
```diff
@@ -524,30 +548,22 @@ module.exports = kind(
...
-			}
+		panel.removeClass('transitioning');
 
-			if (this._previousPanel) {
-				this._previousPanel.removeClass('transitioning');
-				this._previousPanel.deactivated && this._previousPanel.deactivated();
-			}
-		}
+		asyncMethod(function () {
+			panel.postTransition();
```
ryanjduffy: Noticed in the moonstone sample that this line can cause a runtime exception when a `transitionend` event bubbles to this control from something other than a panel (the ScrollThumb in the sample's case). We need to guard against that somehow here (and the prior line to remove the transitioning class). An `instanceof` check would probably work but it may be sufficient to move these into the `if` block above (assuming that makes logical sense here). An (better, i think) alternative would be to fire a custom event from LightPanel when `transitionend` fires and put the responsibility there. Thoughts?

aarontam: For posterity + reference, a mild reworking to utilize states instead of events:

https://github.com/enyojs/enyo/compare/ENYO-1981-aarontam...ENYO-1981-states-aarontam
https://github.com/enyojs/moonstone/compare/ENYO-1981-aarontam...ENYO-1981-states-aarontam

PR #1226, opened 2015-07-07
Fixes ENYO-2032

ryanjduffy:
## Issue
Calling `focus()` on a control within the floating layer causes the browser to scroll it into view which can break things like moonstone/Popup that are designed to animate into view.

## Fix
Prevent scrolling of the floating layer

### DIFF
--- a/lib/floatingLayer.js
+++ b/lib/floatingLayer.js
@@ -6,6 +6,7 @@
 require('enyo');
 
 var
+	dispatcher = require('./dispatcher'),
 	kind = require('./kind'),
 	platform = require('./platform');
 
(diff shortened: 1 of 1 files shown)


DavidUM:
Hi @ryanjduffy , Thank you for your fix.
I didn't know this behavior. So interesting.
For further understanding, could you recommend an article, blog or manual which explain this behavior? 
I'd like to fully understand it.
Thanks!

seunghoh:
@DavidUM come to me, I'll explain in detail :)
This is because we call focus() 'out of view' component on accessibility condition. This depends on web engine - FF doesn't occur, Safari and Chrome occur. From blink team's analysis, web engine seems not to consider the position after transform animation.
There are a lot of related Q&As - you can search with "prevent scrollIntoView". https://forum.jquery.com/topic/chrome-bug-or-how-do-i-prevent-a-form-field-to-scroll-the-container-when-focused

PR #1227, opened 2015-07-09
Fixes ENYO-2032

ryanjduffy:
Because there are a few places that can exhibit this bug (all thus far
demonstrated by programmatic focus from Spotlight), we've moved this
logic in AccessibilitySupport which is mixed into Control so it's more
broadly available.

### DIFF
--- a/lib/AccessibilitySupport/AccessibilitySupport.js
+++ b/lib/AccessibilitySupport/AccessibilitySupport.js
@@ -1,5 +1,22 @@
 var
+	dispatcher = require('../dispatcher'),
 	kind = require('../kind');
+
+
+/**
+* Prevents browser-initiated scrolling contained controls into view when those controls are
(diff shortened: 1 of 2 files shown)


lib/AccessibilitySupport/AccessibilitySupport.js
```diff
@@ -80,6 +97,16 @@ module.exports = {
 	accessibilityDisabled: false,
 
 	/**
+	* When true, `onscroll` events will be observed and scrolling prevented by resetting the
+	* `scrollTop` and `scrollLeft` of the node. This prevents inadvertent
```
aarontam: I think you were in the middle of a thought here. :smile:

ryanjduffy: documentation fail ... (sadpanda)

PR #1228, opened 2015-07-15
ENYO-2062: Corrected fullscreen pseudo selector to match W3 spec

Djspaceg:
Also updated IE's fullscreen pseudo-selector.
More details: https://developer.mozilla.org/en-US/docs/Web/CSS/:fullscreen


ryanjduffy:
Should this use same approach as enyojs/moonstone#2190 both for consistency and end dev usage (e.g. they wanted to apply fullscreen styles without worrying about the vendor prefixes or the bug we're fixing here)?

Djspaceg:
Perhaps! I'll give it a spin.

PR #1230, opened 2015-07-21
ENYO-1013: Always attempt to update the source, regardless of generated state.

aarontam:
### Issue
When setting the `sizing` property at component declaration time, the source is not properly set to be a background image.

### Fix
We move the call to `srcChanged` outside the guard for the generated state.


lishammel:

PR #1232, opened 2015-07-23
ENYO-2145: More robust value parsing.

aarontam:
### Issue
The LESS API is fairly lacking and the initial pass of the resolution-independence plugin felt more like fumbling around in the dark. Now that we have a better idea of the API's behavior, we can take a step back and rework the parsing logic to more broadly handle different cases we will encounter.

### Fix
The parsing has been reworked to handle a variety of cases by recursively parsing each "value" until it can be resolved via conversion or is a string that does not require conversion. This is a version of https://github.com/enyojs/less-plugin-resolution-independence/pull/1 without tests. It would require some rejiggering of the plugin in order to make it effectively testable, and because the 2.6.x version was the high-priority need, we can either spend some more time on the 2.5.x version to make it testable (and integrate later), or move forward without tests for this version as it is effectively a dead-end branch.


lishammel:

PR #1240, opened 2015-07-29
ENYO-2111: ExpandablePicker: Grouped Picker Bounces on Initial Selection

iKuzko:
ENYO-2111: ExpandablePicker: Grouped Picker Bounces on Initial Selection


sakhnik:
Combine the two commits.

PR #1241, opened 2015-07-31
ENYO-2044: Add aria role and tabindex for reading image alt

kbs12e:
## Issue 
Accessibility: screen reader does not read 'alt' value of image

## Cause
Screen reader can read text of focusable component. However, enyo Image uses img tag or div placeholder and these both are not focusable, so we need to add tabindex to enyo.Image for accessibility feature. In additon, aria img role also is necessary because enyo.Image's sizing API uses div tag.

## Fix
add tabindex="0" and img role to enyo Image.


rylan:
You do not normally set tabindex on images, how are you testing this? (reader and browser combo?)
Navigating through the page content with the arrow keys not picking up the image? 
Have you tried just setting the aria-label on an image?

kbs12e:
The tabindex and role is added when accessibility of option is only true for blind. accessibility value is default false.

rylan:
You should not assume that a11y is only used for blind users, setting a tabindex effects keyboard accessibility. Generally you only set the tabindex on an image if you want to change the order in which an image is read from its structure in the page. Because by setting the tabindex you have now made the image a part of the tab navigation of the page.

How are you testing your changes?

PR #1242, opened 2015-07-31
ENYO-2129 : add aria-pressed to child Item of Group

kbs12e:
## Issue 
Accessibility: Do not read the active status of the grouped buttons

## Cause
Button wrapped using enyo Group does not have aria-pressed attribute, which is neccessary for reading own status. However, If Button is not child of Group, Button should not have aria-pressed attribute since it is used in ToggleButton for reading toggle status, so we should add only aria-pressed attribute when button is child of group. 

## Fix
Add aria-pressed attribute when button is child of Group.


rylan:
You should be using role radiogroup and role radio here, not button.

kbs12e:
accessibility is refactoring, so this PR needs more changes. Thus , I will close this PR and create new PR.

PR #1243, opened 2015-07-31
ENYO-2198: Added support for new `groupable` option

Djspaceg:
`Groupable:false` will ignore the fact that it's inside a group. 😥
Evidently also much trailing whitespace cleanup.

Enyo-DCO-1.1-Signed-off-by: Blake Stephens blake.stephens@lge.com


lishammel:

PR #1245, opened 2015-08-03
Initial commit

anishramesan:
Initial commit for "animation" with enyo components.

Enyo-DCO-1.1-Signed-off-by: Anish Ramesan anish.ramesan@lge.com

### DIFF
--- a/lib/UiComponent.js
+++ b/lib/UiComponent.js
@@ -8,7 +8,8 @@ require('enyo');
 var
 	kind = require('./kind'),
 	utils = require('./utils'),
-	master = require('./master');
+	master = require('./master'),
+	AnimationSupport = require('./animation/AnimationSupport');
 
 var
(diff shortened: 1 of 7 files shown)


Rayraz:
Very cool approach for a more powerful and centralized implementation of animations. Would it be possible to add the ability to make the keyframe of an animated property computable?

Another thought, other than only using tweens, it would be cool to have another animation strategy that gives an animated property a momentum and treat the animations' keyframe as an attractor that applies a force to the property. Add some friction and you can create animations with satisfying wiggles or let a control move more smoothly if a properties target value is altered mid-animation.

If you remove the need for a keyframe and simply apply force to a property you could also easily let objects gradually grind to a hold on their current momentum after a drag or scroll interaction from users.

Lots of interesting possibilities here!

PR #1247, opened 2015-08-07
ENYO-2205 Initialize selected items when collection changed

webOS101:
* Issue *
Selection handling in DataRepeater/DataList was inconsistent. Selections would not be properly tracked with DataList and selections were not reset when changing collections.

* Solution *
If a `selectionProperty` is defined, filter out selected items to initialize the internal `_selection` property.

### DIFF
--- a/lib/DataRepeater.js
+++ b/lib/DataRepeater.js
@@ -16,6 +16,14 @@ function at (idx) {
 	return this[idx];
 }
 
+function arrayFilter (record) {
+	return record[this.selectionProperty];
+}
+
+function modelFilter (record) {
+	return record.get(this.selectionProperty);
+}
+
 /**
 * {@link module:enyo/DataRepeater~DataRepeater} iterates over the items in an {@link module:enyo/Collection~Collection} to
 * repeatedly render and synchronize records (instances of {@link module:enyo/Model~Model}) to its
@@ -60,9 +68,9 @@ var DataRepeater = module.exports = kind(
 	* selection and deselection of a single item at a time. The 'multi' selection mode allows
 	* multiple children to be selected simultaneously, while the 'group' selection mode allows
 	* group-selection behavior such that only one child may be selected at a time and, once a
-	* child is selected, it cannot be deselected via user input. The child may still be 
+	* child is selected, it cannot be deselected via user input. The child may still be
 	* deselected via the selection API methods.
-	* 
+	*
 	* @type {String}
 	* @default 'single'
 	* @public
@@ -454,7 +462,8 @@ var DataRepeater = module.exports = kind(
 	* @private
 	*/
 	initCollection: function (c, p) {
-		var e;
+		var e, filter, isArray = c && c instanceof Array;
+
 		if (c && c.addListener) {
 			for (e in this._handlers) {
 				c.addListener(e, this._handlers[e]);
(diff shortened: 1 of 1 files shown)


lib/DataRepeater.js
```diff
@@ -454,7 +454,7 @@ var DataRepeater = module.exports = kind(
 	* @private
 	*/
 	initCollection: function (c, p) {
-		var e;
+		var e, isArry = false;
```
aarontam: Can we make this `isArray` or `isArr`? :smile:

webOS101:
Updated variable name and discovered test is not very good so updated the test.

lib/DataRepeater.js
```diff
@@ -471,6 +473,14 @@ var DataRepeater = module.exports = kind(
 				p.removeListener(e, this._handlers[e]);
 			}
 		}
+		if(c && this.selectionProperty) {
+			var filter = isArray ?
+				function(record) { return record[this.selectionProperty]; } :
+				function(record) { return record.get(this.selectionProperty); };
```
rajyavardhanp: functions should be bound to the scope to identify the scope's properties inside I guess.

lib/DataRepeater.js
```diff
@@ -463,6 +463,8 @@ var DataRepeater = module.exports = kind(
 		// Decorate native JS array with at() so that we can
 		// access members of our dataset consistently, regardless
 		// of whether our data is in an array or a Collection
+		// TODO: Document that we add a method to an array or use
```
ryanjduffy: Rather than documenting this, let's just fix it.

`Object.defineProperty(c, 'at', {value: at, enumerable: false})`

lib/DataRepeater.js
```diff
@@ -215,9 +215,9 @@ var DataRepeater = module.exports = kind(
 	create: kind.inherit(function (sup) {
 		return function () {
 			sup.apply(this, arguments);
+			this.selectionType = this.multipleSelection ? (this.groupSelection ? 'group' : 'multi') : this.selectionType;
 			this.collectionChanged();
 			// Converting deprecated selection properties to our current selection API
-			this.selectionType = this.multipleSelection ? (this.groupSelection ? 'group' : 'multi') : this.selectionType;
```
ryanjduffy: Why was this moved up?

lib/DataRepeater.js
```diff
@@ -471,6 +473,17 @@ var DataRepeater = module.exports = kind(
 				p.removeListener(e, this._handlers[e]);
 			}
 		}
+		if(c && this.selectionProperty) {
```
ryanjduffy: Space between `if` and `(`

lib/DataRepeater.js
```diff
@@ -471,6 +473,17 @@ var DataRepeater = module.exports = kind(
...
+			var selectionProperty = this.selectionProperty,
```
ryanjduffy: While we're at it, let's move these `var`s up to the top of the method. Might consider defining these methods at the top of the module (like `at()`).

lib/DataRepeater.js
webOS101: Because selectionType needs to be set before the collectionChanged method is invoked.

lib/DataRepeater.js
webOS101: Good idea.  I have done so, and moved (and renamed) the local shadow of selectionProperty

lib/DataRepeater.js
aarontam: Should we move the accompanying comment, as well, then?

webOS101: (rageface)  yes

aarontam: Actually, does not seem like we need to move this line - going to move it back and save the comment from having to pack its things.

lib/DataRepeater.js
```diff
@@ -464,13 +476,20 @@ var DataRepeater = module.exports = kind(
...
-			c.at = at.bind(c);
+			Object.defineProperty(c, 'at', {value: at, enumerable: false});
 		}
 		if (p && p.removeListener) {
 			for (e in this._handlers) {
 				p.removeListener(e, this._handlers[e]);
 			}
 		}
+		if (c && this.selectionProperty) {
+			filterSelectionProperty = this.selectionProperty;
```
aarontam: Going to make a small tweak so that we don't need to maintain this var, and instead will pass the context to the `filter` method.

PR #1249, opened 2015-08-07
ENYO-2257 DataList/DataRepeater does not update selection with POJOs

webOS101:
## Issue
POJO selection not working because bindings do not update when we change the underlying selection property

## Solution
Call the `syncBindings()` method on the row when we update the selection property

### DIFF
--- a/src/DataRepeater.js
+++ b/src/DataRepeater.js
@@ -630,6 +630,7 @@ var DataRepeater = module.exports = kind(
 			}
 			else {
 				model[p] = select;
+				if(c) c.syncBindings({force: true, all: true});
 			}
 		}
 		this.notifyObservers('selected');


lib/DataRepeater.js
```diff
@@ -615,6 +615,7 @@ var DataRepeater = module.exports = kind(
 			}
 			else {
 				model[p] = select;
+				if(c) c.syncBindings({force: true, all: true});
```
clinuz: Is this really the only way to do this? Won't it automatically sync when not using POJO's? So doing this blindly with `force: true` seems redundant and wasteful.

webOS101: This only executes on POJOs

clinuz: Ah, yup!

PR #1250, opened 2015-08-10
Web Storage can now use localStorage or sessionStorage

Zectbumo:
Enyo-DCO-1.1-Signed-off-by: Alfred Morgan <alfred@54.org>


dmikey:
I like this line of thought, I wonder if this shouldn't be abstracted out a little more though, localStorage and  sessionStorage can be two kinds that inherit from a webStorage object. I think that feels more enyoisque than setting an option, it also allows us to cleanly support the old API if anyone is using that. Thoughts?

Zectbumo:
I also felt that the option was not enyoisque. I feel abstraction is a bit overkill though. It's a basic web storage interface and one option to change it. Two subclasses and a base class is a bit much to simply switch web stores. Abstraction would be nice if it could be used outside of web storage but the point here is to support the web storage api so I don't see any value added. It would make more sense to me if I found all my Web Storage needs in one class/file. Is there any other enyoisque method you can recommend to achieve this?

dmikey:
I imagine all the basic structure would appear in one file, instead of using an enyo object maybe it would be cleaner to make those methods vanilla JS, and then us it as a mixin into LocalStorage and SessionStorage kinds. 

-- WebStorageSupport.js
-- SessionStorage.js mixins[webstoragesupport]
-- LocalStorage.js mixins[webstoragesupport]


In current master this can all live in 1 file, for our 2.6 modular branch it would be in the 3 files like above.

Zectbumo:
Care to write up a working version of that?

lishammel:
After additional internal discussion, going to close this PR out.

PR #1253, opened 2015-08-20
ENYO-2258-Scroller in list action is not showed

rajyavardhanp:
Issue: Scroller is '0' height as its parent is reflowed while its in
hidden state
Fix: While reflowing, the particular state is remembered and while its
showing based on the state the control is reflowed again.
Enyo-DCO-1.1-Signed-off-by: Rajyavardhan P <rajyavardhan.p@lge.com>

### DIFF
--- a/lib/UiComponent.js
+++ b/lib/UiComponent.js
@@ -619,7 +619,7 @@ var UiComponent = module.exports = kind(
 	reflow: function () {
 		if (this.layout) {
 			this.layout.reflow();
-			this._hideReflow = this.showing? false: true;
+			this._hiddenReflow = this.showing ? false : true;
 		}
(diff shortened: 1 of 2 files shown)


lib/UiComponent.js
```diff
@@ -619,6 +619,7 @@ var UiComponent = module.exports = kind(
 	reflow: function () {
 		if (this.layout) {
 			this.layout.reflow();
+			this._hideReflow = this.showing? false: true;
```
dragongears: There should be a space before the question mark and colon.

lishammel:
After discussion with Rajyavardhan, closing out as jira associated with this has been resolved another way.

PR #1257, opened 2015-08-24
Enyo 2368 Implement history back supporting when app does not use history API

DavidUM:
Issue
-------
When we'd like to use RC's back key, `disableBackHistoryAPI = true`
But in this case, enyo.History does not support back key

Cause
---------
Refactored History does not consider `disableBackHistoryAPI = true'

Fix
----
Add a logic path when `disabledBackHistoryAPI = true`


ryanjduffy:
Closing in favor of #1258

PR #1271, opened 2015-09-04
ENYO-2452-rajyavardhan-Unnecessary border around placeholder if image…

rajyavardhanp:
… fails to load

Issue: The behavior is rather normal as enyo.Image makes the 'src:
null', if the image fails to load. So browser draws a border around the
size of the image when there is no 'src' attribute or 'src: null'
Fix: Instead of making the src: null when the image doesn't load, the
source is set with a transparent image.
Enyo-DCO-1.1-Signed-off-by: Rajyavardhan P <rajyavardhan.p@lge.com>

### DIFF
--- a/src/Image/Image.js
+++ b/src/Image/Image.js
@@ -182,7 +182,7 @@ module.exports = kind(
 
 	/**
 	* This image is used to handle the border around placeholder, if the original image
-	* to load
+	* fails to load
 	*
 	* @private
 	*/
--- a/src/Image/Image.js
+++ b/src/Image/Image.js
@@ -186,7 +186,16 @@ module.exports = kind(
 	*
 	* @private
 	*/
-	_transparentImg: 'data:image/gif;base64,R0lGODlhAQABAAAAACH5BAEKAAEALAAAAAABAAEAAAICTAEAOw==',
+	_transparentImg:
+		'data:image/svg+xml;base64,PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciI' +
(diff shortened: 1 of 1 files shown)


src/Image/Image.js
```diff
@@ -276,7 +284,7 @@ module.exports = kind(
 	*/
 	handleLoad: function () {
 		if (!this.sizing && this.placeholder) {
-			this.applyStyle('background-image', null);
+			this._errorHandling? this._errorHandling = false : this.applyStyle('background-image', null);
```
ryanjduffy: I'd rather see an if/else block here rather than the ternary operator for clarity.

src/Image/Image.js
```diff
@@ -285,7 +293,8 @@ module.exports = kind(
 	*/
 	handleError: function () {
 		if (this.placeholder) {
-			this.set('_src', null);
+			this.set('src', this._transparentImg);
```
ryanjduffy: We shouldn't overwrite the user's value of `src`. That's why we have the `_src` member.

src/Image/Image.js
```diff
@@ -181,6 +181,14 @@ module.exports = kind(
 	_src: null,
 
 	/**
+	* This image is used to handle the border around placeholder, if the original image
+	* to load
+	*
+	* @private
+	*/
+	_transparentImg: 'data:image/gif;base64,R0lGODlhAQABAAAAACH5BAEKAAEALAAAAAABAAEAAAICTAEAOw==',
```
ryanjduffy: @Djspaceg - is there an SVG version we'd prefer here?

Djspaceg: We do not have an SVG version, but we should make one. Better for scalability.

clinuz:
Closed as "Not A Bug".

PR #1272, opened 2015-09-04
Fixes ENYO-1994

ryanjduffy:
notify observers of index when using LightPanels.animateTo

Without the notification, the aria observer cannot update the role of
the active LightPanel.

### DIFF
--- a/src/LightPanels/LightPanels.js
+++ b/src/LightPanels/LightPanels.js
@@ -192,7 +192,6 @@ module.exports = kind(
 			this.updateTransforms();
 			this.orientationChanged();
 			this.directionChanged();
-			this.indexChanged();
 		};
 	}),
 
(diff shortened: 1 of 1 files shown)


src/LightPanels/LightPanels.js
```diff
@@ -192,7 +192,6 @@ module.exports = kind(
 			this.updateTransforms();
 			this.orientationChanged();
 			this.directionChanged();
-			this.indexChanged();
```
aarontam: I think we may need something like this in the future, to support specifying a starting index. This isn't really a common use case right now, so I'll open a ticket for follow-on work, and we can figure out a way to work around this.

ryanjduffy: It probably should have changed to a call to `setupTransitions()` rather than removing it since that was effectively what was happening before.

PR #1273, opened 2015-09-05
ENYO-2465: Allow for declared component block and initial index.

aarontam:
### Issue
There were some issues with using a declared component block for `LightPanels`. Additionally, we did not support setting of an initial index in this case.

### Fix
We now only set the index to `-1` if we have no panels after creation, otherwise we default to the specified value of `index`, which defaults to `0` unless overridden. Additionally, a tweak was made such that all inactive panels are offscreen until activated - I decided to change the position versus hiding the panels, as that could have some adverse/unexpected effects especially for the panel's child components, but am definitely open to further discussion or thoughts on this. Lastly, cleaned-up the styles to be more efficient and remove the use of an unnecessary class.


src/LightPanels/LightPanels.js
```diff
@@ -183,17 +183,28 @@ module.exports = kind(
...
+		Control.prototype.create.apply(this, arguments);
+		this.updateTransforms();
+		this.orientationChanged();
+		this.directionChanged();
+		this.setupTransitions();
+	},
+
+	initComponents: function () {
+		Control.prototype.initComponents.apply(this, arguments);
+		if (!this.components || !this.components.length) this.index = -1;
```
ryanjduffy: The assumption here that LightPanel instances would be declared via components isn't necessarily valid. If you create a subkind of LightPanels which included it's own internal panels, those wouldn't be declared in `components`. Suggest we move this to `create()` (so there's a logical opportunity to create panels by a subkind) and use `this.getPanels()` instead of the design-time `components` property.

PR #1281, opened 2015-09-14
ENYO-2514: VDR: Always refresh immediately

graynorton:
VirtualDataRepeater has, until now, routinely made refresh() an
async operation. This behavior was inherited (via copy/paste) from
DataRepeater when VDR was first created.

It's not apparent that making refresh() async is necessary or
useful for VDR, and the async operation is causing problems in at
least one instance -- specifically, when the data underlying a
NewDataList is modified while the list is hidden, the list isn't
refreshed until the list is reshown, and at that time the "old"
list contents flash momentarily before being refreshed.

We address this issue by making refresh() synchronous for VDR and
its subkinds.

### DIFF
--- a/src/VirtualDataRepeater.js
+++ b/src/VirtualDataRepeater.js
@@ -70,18 +70,7 @@ module.exports = kind({
 
 		this.stabilizeExtent();
 
-		var refresh = this.bindSafely(function() {
-			this.doIt();
-		});
-
-		// refresh is used as the event handler for
-		// collection resets so checking for truthy isn't
-		// enough. it must be true.
-		if(immediate === true) {
-			refresh();
-		} else {
-			this.startJob('refreshing', refresh, 16);
-		}
+		this.doIt();
 	},
 
 	childForIndex: function(idx) {


src/VirtualDataRepeater.js
```diff
@@ -70,18 +70,7 @@ module.exports = kind({
...
-
-		// refresh is used as the event handler for
-		// collection resets so checking for truthy isn't
-		// enough. it must be true.
-		if(immediate === true) {
-			refresh();
-		} else {
-			this.startJob('refreshing', refresh, 16);
-		}
+		this.doIt();
```
clinuz: This change makes the `immediate` parameter unused and deprecated. Also, does it actually need to refresh when hidden? If lots of changes are happening in succession this would waste a lot of time. What if refreshing were handled much like the new reflow-when-not-showing system forcing it to update when showing changes? I think this is another reason why it was async to begin with in DataList because if lots of sequential updates were happening to the collection we didn't need to try and update the list over and over for each change...

graynorton: Yeah, this effectively means that VDR ignores the `immediate` parameter (although DataRepeater itself still respects it).

Currently there's logic in NewDataList that defers refresh, along the same lines as the reflow-when-showing solution. We could consider moving that to VDR instead.

PR #1283, opened 2015-09-15
ENYO-2258: Defer reflow when hidden, do it when shown

graynorton:
Reflowing a Control when it's hidden generally doesn't make sense,
since a) it's wasteful to spend cycles laying out something that
can't be seen; and b) the layout may depend on measuring the
Control's children, and they can't be accurately measured if
they're not showing.

We therefore defer reflow in the case where we're hidden, and
wait until we're shown.

In the case where we are explicitly hidden (i.e., `showing` is
`false`), we defer immediately. We can't inexpensively detect the
case where we're showing but an ancestor is hidden, so to handle
that case we rely on the layout to do its measurements and let
us know (by returning `true`) if we need to reflow again the next
time we are shown.

### DIFF
--- a/src/Control/Control.js
+++ b/src/Control/Control.js
@@ -882,7 +882,25 @@ var Control = module.exports = kind(
 	* @private
 	*/
 	showingChangedHandler: function (sender, event) {
+		// If we have deferred a reflow, do it now...
+		if (this.showing && this._needsReflow) {
+			this.reflow();
+		}
+
+		// Then propagate `onShowingChanged` if appropriate
 		return sender === this ? false : !this.showing;
+	},
+
+	/**
+	* Overriding reflow() so that we can take `showing` into
+	* account and defer reflowing accordingly.
+	*
+	* @private
+	*/
+	reflow: function () {
+		if (this.layout) {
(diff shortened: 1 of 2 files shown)


Rayraz:
I think there is an easy way to make cases where the ancestor is hidden more readily detactable.

Whenever a Control changes it's showing state, it waterfalls an `onShowingChanged` event to it's descendants. Currently, this event is not propagated when it runs into a hidden descendant.

Instead of stopping propagation and doing nothing, you could send a new event that lets descendants know they're hidden instead.

For example:
```javascript
showingChangedHandler: function(inSender, inEvent) {
    if (inEvent.showing && !this.getShowing()) {
        // Send new accurate event.
        this.sendShowingChangedEvent();
        return true;
    } else {
        // Proceed waterfalling unchanged
        return false;
    }
}
```

This probably requires a little coordination to work well in combination with the `absoluteShowing` property, but it should work. I've got an imperfect - but considerably better than nothing - version running in enyo 2.4 that's doing a pretty decent job.

By the way, you could use a task queue to defer all kinds of actions when a component isn't showing. Aside from reflows, animations could also be deferred and/or fast-forwarded.

PR #1286, opened 2015-09-16
ENYO-2536: Guard against initialization edge cases

aarontam:
### Issue
When initially populating a `LightPanels` kind, if `pushPanels` is used, by default the index is set to be the index of the last panel pushed. If we attempt to set the index to a different value, immediately after this operation, the `ontransitionend` event for the must recent index change does not always fire, especially on slower hardware, as the rAF operation effectively batched both the transition setup and the transition application as the 16ms timeout is not sufficient during what is presumably some heavy load (executing the previous transition), resulting in no transition event occurring at all.

### Fix
To counter this, we separate out the two operations into different rAF requests, and kick off the transition application rAF *after* the transition setup rAF. This way there is a redraw in-between and allows for the transition to always be applied. Also, this allows us to drop the somewhat-arbitrary 16ms timeout we had been using to semi-guarantee that a redraw had occurred between these operations.

### DIFF
--- a/src/LightPanels/LightPanels.js
+++ b/src/LightPanels/LightPanels.js
@@ -8,6 +8,7 @@ require('enyo');
 var
 	kind = require('../kind'),
 	dom = require('../dom'),
+	animation = require('../animation'),
 	asyncMethod = require('../utils').asyncMethod;
 
 var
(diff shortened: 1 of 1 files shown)


aarontam:
I think we have to respect a value of 0 for `this._previousIndex`.

PR #1298, opened 2015-10-01
Fixes ENYO-2619

ryanjduffy:
add kind() method to Kind constructors

### DIFF
--- a/src/kind.js
+++ b/src/kind.js
@@ -83,6 +83,11 @@ var kind = exports = module.exports = function (props) {
 	utils.forEach(kind.features, function(fn){ fn(ctor, props); });
 	
 	if (name) kindCtors[name] = ctor;
+
+	ctor.kind = function (props) {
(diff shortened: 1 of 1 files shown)


src/kind.js
```diff
@@ -83,6 +83,11 @@ var kind = exports = module.exports = function (props) {
 	utils.forEach(kind.features, function(fn){ fn(ctor, props); });
 	
 	if (name) kindCtors[name] = ctor;
+
+	ctor.kind = function (props) {
```
clinuz: Shouldn't this be part of the static handling? As of this code, it would be creating a new function instance for every _kind_ but they can easily inherit the same method attached to different contexts. It would be static so it would look at `this` as the constructor in question as opposed to a memoized reference. This would also decrease stack frames/context switching when executed. JM2C!

ryanjduffy: Good call!

PR #1301, opened 2015-10-05
ENYO-2646 Move contructor to create

webOS101:

### DIFF
--- a/src/DataGridList/DataGridList.js
+++ b/src/DataGridList/DataGridList.js
@@ -80,7 +80,7 @@ var DataGridList = module.exports = kind(
 	* @method
 	* @private
 	*/
-	constructor: kind.inherit(function (sup) {
+	create: kind.inherit(function (sup) {
 		return function () {
 			var o = this.orientation;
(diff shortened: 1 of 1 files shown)


src/DataGridList/DataGridList.js
```diff
@@ -80,7 +80,7 @@ var DataGridList = module.exports = kind(
 	* @method
 	* @private
 	*/
-	constructor: kind.inherit(function (sup) {
+	create: kind.inherit(function (sup) {
```
aarontam: I realize it was part of the original code, but we should get rid of the `return` on L92.

aarontam:
I think we will also need to change this to `create`: https://github.com/enyojs/moonstone/blob/2.6.0-dev/src/DataGridList/DataGridList.js#L94.

src/DataGridList/DataGridList.js
webOS101: Might be a good idea to always return the result, though I don't think it matters for create.

PR #1308, opened 2015-10-14
ENYO-2698: Reset queue and history instead of enqueueing drop when clearing history

viodragon2:
### Issue
When `clear()` is called when one or more `drop()` is being queued, length of `_history` array does not reflect the correct number of history entities to clear and ends up dropping more than it needs to. This results in splicing the `_history` array with negative starting index when processing the enqueued pop.

### Fix
Clear function would remove all entities in `_history` as well as the `_queue`.

### DIFF
--- a/src/History.js
+++ b/src/History.js
@@ -182,9 +182,9 @@ var EnyoHistory = module.exports = kind.singleton(
 	* @public
 	*/
 	clear: function () {
-		_queue.length = 0;
+		_queue.splice(0, _queue.length);
+		_history.splice(0, _history.length);
 		_popQueueCount = 0;
-		_history.length = 0;
 		_pushQueued = false;
 	},
 
--- a/src/History.js
+++ b/src/History.js
@@ -183,6 +183,9 @@ var EnyoHistory = module.exports = kind.singleton(
 	*/
 	clear: function () {
 		var len = _history.length;
+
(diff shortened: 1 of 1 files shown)


src/History.js
```diff
@@ -175,17 +175,17 @@ var EnyoHistory = module.exports = kind.singleton(
...
 	* reset to the state when this module started tracking the history.
 	*
 	* @public
 	*/
 	clear: function () {
-		var len = _history.length;
-		if (len) {
-			this.drop(len);
-		}
+		_queue.length = 0;
```
clinuz: I think we need to run through and pop off the array to ensure that we aren't inadvertently keeping orphaned objects in memory. If the array were merely simple types, this would be an ok performance hack.

src/History.js
```diff
@@ -175,17 +175,17 @@ var EnyoHistory = module.exports = kind.singleton(
...
 	*
 	* @public
 	*/
 	clear: function () {
-		var len = _history.length;
-		if (len) {
-			this.drop(len);
-		}
+		_queue.splice(0, _queue.length);
+		_history.splice(0, _history.length);
```
dmikey: Wouldn't it be better to actually set the length to zero, or the history array to [] , rather than splicing? This would perform a shallow copy, and the GC will need to wait for the closure to complete before releasing ref to these objects still?

PR #1310, opened 2015-10-16
ENYO-2697: Stop job when clearing to ensure drop handlePop does not g…

viodragon2:
…et executed

### DIFF
--- a/src/History.js
+++ b/src/History.js
@@ -102,12 +102,7 @@ var EnyoHistory = module.exports = kind.singleton(
 	*/
 	enabledChanged: function () {
 		// reset private members
-		_history = [];
-		_queue = [];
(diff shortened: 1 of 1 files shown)


src/History.js
```diff
@@ -186,6 +186,7 @@ var EnyoHistory = module.exports = kind.singleton(
 		_history.splice(0, _history.length);
 		_popQueueCount = 0;
 		_pushQueued = false;
+		this.stopJob('history.go');
```
viodragon2: This just becomes the same as enabledChanged(). I wonder if enabledChanged should simply call this.clear() instead and add _processing = false;

webOS101: Good catch!

PR #1314, opened 2015-10-21
ENYO-2719: Remove panel in reverse order when replacing panels

viodragon2:
## Issue 
When replaceAt() goes through the loop to remove panels, removePanel changes the length and entries of panels array. This will skip every even numbered index of panels that needs to be removed.

## Fix 
Remove panel in reverse order.

### DIFF
--- a/src/LightPanels/LightPanels.js
+++ b/src/LightPanels/LightPanels.js
@@ -531,7 +531,7 @@ module.exports = kind(
 		commonInfo = {addBefore: insertBefore};
 
 		// remove existing panels
-		for (idx = start; idx < end; idx++) {
+		for (idx = end - 1; idx >= start; idx--) {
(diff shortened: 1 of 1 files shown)


src/LightPanels/LightPanels.js
```diff
@@ -531,7 +531,7 @@ module.exports = kind(
 		commonInfo = {addBefore: insertBefore};
 
 		// remove existing panels
-		for (idx = start; idx < end; idx++) {
+		for (idx = end - 1; idx >= start; idx--) {
```
dragongears: If count is 0 (or < 0) then idx is going to start out < start and go down from there. Do we want to guard against this?

webOS101: Not a concern, it will be >= start and exit the loop immediately.

PR #1315, opened 2015-10-23
ENYO-2533, ENYO-2696, ENYO-2514: Overhaul refresh / reset in VDR, NewDataList

graynorton:
This change moves the logic governing deferred refresh() and reset() behavior up the inheritance chain from NewDataList to VirtualDataRepeater and reintroduces async refresh for performance reasons.

In the process, it addresses ENYO-2533 and ENYO-2696, and replace the original fix for ENYO-2514.

See the notes on the individual commits for more detail.

### DIFF
--- a/src/NewDataList.js
+++ b/src/NewDataList.js
@@ -323,57 +323,26 @@ module.exports = kind({
 	/**
 	* @private
 	*/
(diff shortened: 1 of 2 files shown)


webOS101:
That's what I get for reviewing a commit at a time.  All clear now!

PR #1322, opened 2015-11-04
Fixes ENYO-2699

ryanjduffy:
A prior change updated clear() to only reset the internal history so
this change corrects clear() to both reset the internal history and
keep the browser history in sync.


viodragon2:
LGTM!

PR #1324, opened 2015-11-05
after tearDownRender, render does not happen

suhlee:
Cause
--------
after tearDownRender, render does not happen, because this and control is different

Fix
----
reverted ENYO-2744 only HTMLStringDelgate.js

Enyo-DCO-1.1-Signed-off-by: Suhyung Lee suhyung2.lee@lge.com

### DIFF
--- a/src/HTMLStringDelegate.js
+++ b/src/HTMLStringDelegate.js
@@ -120,7 +120,7 @@ module.exports = {
 	* @private
 	*/
 	renderContent: function (control) {
-		if (control.generated) control.teardownChildren();
+		if (control.generated) this.teardownChildren(control);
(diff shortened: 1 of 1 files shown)


graynorton:
Thanks, @suhlee!

I'm going to close this in favor of a different approach. The changes to HTMLStringDelegate were actually required to fix the original issue ENYO-2744, so simply reverting them won't help.

The more specific problem was that the `cache` argument to `teardownChildren()` was not being preserved in the new flow between HTMLStringDelegate and Control. One way to fix it would have been to simply make sure we passed the argument properly between the two of them, but I opted instead for a different option -- rather than promote the overriding of `teardownChildren()`, we'll provide an explicit `beforeTeardown()` lifecycle hook that Controls can implement as needed.

PR #1325, opened 2015-11-05
Fixes ENYO-2797

ryanjduffy:
invoke callback if scrolling is not required


ryanjduffy:
Bad idea ...

PR #1333, opened 2015-11-16
ENYO-2816: Fix VDR child indexing logic (again)

graynorton:
It has taken a few tries, but I think this finally gets us to the
bullet-proof child-indexing solution we've been chasing.

We simply rebuild the index from scratch each time we run the
virtualization logic in doIt().


Djspaceg:
Gotta love that `dolt` function.

PR #1335, opened 2015-11-16
Fixes PLAT-12048 add dismiss event when dragging VM

ryanjduffy:
Also add dragging property to distinguish between dismiss events


ybsung:
It looks good to me. I'd like to merge this PR into viewmanager branch.

PR #1337, opened 2015-11-23
Fixes PLAT-12619

ryanjduffy:
Prevents resetting transform on the dragView which caused a visual jump
when the view reset its position before transitioning offscreen. Resets
dragging when a view is activated and correctly retains a to be
tore down node until dragging completes.


ryanjduffy:
Don't want to target the release branch for this. Will hold in its JIRA branch for now.

PR #1338, opened 2015-11-23
Fixes PLAT-12457

ryanjduffy:
Altered ViewManager.draggable to be either Boolean or a String to
support draggable, flickable, both or neither.

Issue: PLAT-12457


ryanjduffy:
Don't want to target the release branch for this. Will hold in its JIRA branch for now.

PR #1339, opened 2015-11-25
Fixes PLAT-11951

ryanjduffy:
Added opts parameter to several methods to allow one-time activation
configuration (e.g. disabling animation for hiding a floating panel).


ybsung:
It looks find to us. Thanks.

PR #1344, opened 2015-12-15
ENYO-2913: Add queued index support.

aarontam:
### Issue
As an alternative solution to https://github.com/enyojs/enyo/pull/1343, the issue involves consecutive calls to change the index of the panels, such as sequential calls to `animateTo` that occur as part of application flow programmatically, and are not the result of direct user interaction (which already had guards in place for). This will allow for support of the existing app logic, in the cases where multiple, sequential calls to `animateTo` are desired, and the results of the last call are expected to be applied.

### Fix
We have implemented queued index support, such that any number of calls to `animateTo` will all eventually be processed, ending with the results of the last call. There is a known issue with mixing sequential calls to `animateTo` with calls to `set('index')`, as we do not have any kind of checks for the generic setter. We could override the generic `set` and provide some support here, but this seemed like a fairly unnecessary use case to support. That being said, I am open to adding this to fully flesh out this solution, but am omitting it for the time being.

### DIFF
--- a/src/LightPanels/LightPanels.js
+++ b/src/LightPanels/LightPanels.js
@@ -202,6 +202,7 @@ module.exports = kind(
 	*/
 	create: function () {
 		Control.prototype.create.apply(this, arguments);
+		this._queuedIndices = [];
 		this.updateTransforms();
 		this.orientationChanged();
 		this.directionChanged();
@@ -339,10 +340,7 @@ module.exports = kind(
 	* @public
 	*/
 	animateTo: function (index) {
-		var from = this.index;
-		this.index = index;
-		this.notifyObservers('index');
-		this.setupTransitions(from, true);
+		this.queueOrSetIndex(index, true);
 	},
 
 	/**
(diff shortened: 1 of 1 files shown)


Rayraz:
I've ran into similar situations a few times with various components, including a component with very panels-like behavior. It seems to me the main choice to make here is whether an index change should always precede the animation, or whether the animation should always precede the index change.

Personally, I tend to prefer the index is set synchronously, preceding the animation. This seems to be the most sensible solution, because the index is often used for purposes other than animation alone. Animations can then be applied as a reaction to the changed index.

Once this is the case, there are multiple strategies that could be used to handle the animations for quickly changing indexes:
- Fire only one animation at a time, evaluating the need for an additional transition at the end of each animation. (a throttled approach)
- Update the transition endpoint of each panel mid-animation to reflect the current index (can lead to many animating panels at once and cause performance issues).
- Create a secondary throttled index next to the synchronous index and use this throttled index to trigger animations (this approach has worked really well for me and seems conceptually similar to the queued index changes in this pull request).

I don't know if any of this conflicts with LG's internal usage of LightPanels, but I noticed two separate pull requests dealing with this same issue and thought I'd put in my 2ct.

ryanjduffy:
Taking the simpler solution in #1343 for the time being to maintain consistency with the current behavior of `next()` and `previous()`

PR #1345, opened 2015-12-18
ENYO-2917 Document wip modules/kinds

webOS101:

### DIFF
--- a/src/NewAnimator.js
+++ b/src/NewAnimator.js
@@ -23,7 +23,7 @@ var
 * @wip
 * @private
 */
-module.exports = kind.singleton({
+module.exports = kind.singleton(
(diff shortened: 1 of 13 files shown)


src/NewAnimator.js
```diff
@@ -8,7 +15,16 @@ var
...
 
+/**
+* {@link module:enyo/NewAnimator~NewAnimator} is a work-in-progress
+*
+* @class NewAnimator
+* @extends module:enyo/CoreObject~Object
+* @wip
+* @private
+*/
 module.exports = kind.singleton({
```
aarontam: I think we have an extra '{' here. :)

aarontam:
Let's add `enyo/LightPanel`, as well.

PR #1346, opened 2015-12-21
ENYO-2718: Added ShowingTransitionSupport

Djspaceg:
A mixin to make showing/hiding animations easier.


src/ShowingTransitionSupport.js
```diff
@@ -0,0 +1,212 @@
...
+				sup.apply(this, args);
+				if (this.showingDuration && this.generated) {
+					this.set('showingTransitioning', true);
+					// Start transition: Apply a class and start a timer.
+					// When timer finishes, run the exit function,
+					// remove the transitioning class
+					// and add the final-state class
+					this.addClass(this.showingClass);
+					this.startJob('showingTransition', function () {
+						if (this.shownMethod && typeof this.shownMethod == 'string') this[this.shownMethod];	// Run the supplied method.
```
ryanjduffy: the `shownMethod` isn't called here, only referenced. Suggest we use `utils.call(this, this.shownMethod)` instead which  already handles either a string or function. Same for the `this.hiddenMethod` below.

src/ShowingTransitionSupport.js
```diff
@@ -0,0 +1,212 @@
...
+					this.addClass(this.shownClass);
+				}
+			} else {
+				// Reset our state classes, in case we switched mid-stream
+				this.removeClass(this.showingClass);
+				this.removeClass(this.shownClass);
+				if (this.hidingDuration && this.generated) {
+					this.set('showingTransitioning', true);
+					this.addClass(this.hidingClass);
+					this.startJob('showingTransition', this.bindSafely(function () {
```
ryanjduffy: No need to `bindSafely` here since `this.startJob` does that for you.

src/ShowingTransitionSupport.js
```diff
@@ -0,0 +1,212 @@
...
+			sup.apply(this, arguments);
+			this.showingDuration = (this.showingDuration === undefined) ? null      : this.showingDuration;
+			this.hidingDuration  = (this.hidingDuration  === undefined) ? null      : this.hidingDuration;
+			this.shownMethod     = (this.shownMethod     === undefined) ? null      : this.shownMethod;
+			this.hiddenMethod    = (this.hiddenMethod    === undefined) ? null      : this.hiddenMethod;
+			this.shownClass      = (this.shownClass      === undefined) ? null      : this.shownClass;
+			this.hiddenClass     = (this.hiddenClass     === undefined) ? 'hidden'  : this.hiddenClass;
+			this.hidingClass     = (this.hidingClass     === undefined) ? 'hiding'  : this.hidingClass;
+			this.showingClass    = (this.showingClass    === undefined) ? 'showing' : this.showingClass;
+			this.showingChanged();
```
ryanjduffy: What do you think about adding another observer versus overriding `showingChanged()`? It's a bit more overhead but more effectively decouples the behavior of the mixin from the underlying implementation. Also, albeit unlikely, it'd be possible to short circuit this logic if a subkind didn't call its super method if it overrode `showingChanged()` causing some unexpected behavior.

I think there are some benefits to it but curious what you think.

src/ShowingTransitionSupport.js
```diff
@@ -0,0 +1,212 @@
...
+	* @method
+	* @private
+	*/
+	showingChanged: kind.inherit(function (sup) {
+		return function (sender, ev) {
+			var args = arguments;
+
+			// Prepare our visual state
+			this.applyStyle('display', null);
+			this.applyStyle('visibility', null);
```
ryanjduffy: Why are you clearing `visibility` here? Is it a safeguard?

src/ShowingTransitionSupport.js
ryanjduffy: Ah, I see that you're deferring the super call until the callback completes so you have to override showingChanged. I'm hesitant to recommend we take this with the assumption that all impacted showing logic is within an upstream `showingChanged` method. If we've made assumptions anywhere that the change handlers for `showing` were synchronous, we'll have issues here.

I think the idea is solid and the implementation looks good otherwise. We could either mark it as WIP noting the possible issues or hold it until after 2.7 when we can look at `showing` support more broadly. I'm leaning toward the latter since this is a new feature without any consumers.

Thoughts @Djspaceg, @graynorton, @aarontam, @clinuz?

graynorton: I would vote for post-2.7 as well.

Djspaceg: I'd vote for ASAP merging. There's several places I want to use this already.

src/ShowingTransitionSupport.js
Djspaceg: Yep. and it's also used to un-hide the showing cases. It's restored/assigned in the hidden cases, but applied in one paint cycle so no immediate repaint cost.

src/ShowingTransitionSupport.js
Djspaceg: Good `call`. I'll change this in both places.

src/ShowingTransitionSupport.js
Djspaceg: Super. Will simplify.

src/ShowingTransitionSupport.js
aarontam: :eyes:

src/ShowingTransitionSupport.js
webOS101: This needs to be flagged @wip at the very least.

Djspaceg: I'm curious what seems unfinished to everyone about it.

graynorton: My hesitance to merge right away is not based on any specific concerns with this idea or its implementation, but on a) Not yet having had sufficient time to thoroughly experiment with this feature at the framework or application level; and b) The sense that we will want to make other view-management-lifecycle changes in the relatively near future and a desire to think about them holistically before we commit to any new API.

Also, I want to get 2.7 out the door as quickly as possible and with as little risk as possible so that we can start doing some comparatively bold and experimental stuff on master without worrying about near-term releases . With that goal in mind, the bar for what gets into 2.7 is very high -- basically, things go in only if they absolutely must.

All that being said, I would be OK with this going in as WIP, as long as we don't think there's any risk to existing behavior -- that is, if there is no impact on apps unless / until they start using the feature.

Djspaceg: Ah ha. Thanks for the explanation. I was not aware of the current release schedule and thought (based on last I heard) that 2.7 was a year away.
This is indeed only a mixin and has no impact on any other components or apps, unless it's explicitly called out by them.

I'm content with a WIP status for 2.7.

src/ShowingTransitionSupport.js
```diff
@@ -0,0 +1,221 @@
...
+* applied immediately, skipping the transition state classes.
+*
+* An optional method may be supplied to fire at the end of either of the transitions, using
+* `hidingMethod` and/or `showingMethod`.
+*
+* @mixin
+* @wip
+* @public
+*/
+var ShowingTransitionSupport = {
```
aarontam: We should be able to just export this directly via `module.exports = {`

src/ShowingTransitionSupport.js
```diff
@@ -0,0 +1,221 @@
...
+	* @readonly
+	* @private
+	*/
+	name: 'ShowingTransitionSupport',
+
+	/**
+	* A read-only property for checking whether we are in the middle of a transition.
+	*
+	* @type {Boolean}
+	* @default false
```
aarontam: Let's add the `@readonly` pragma here.

src/ShowingTransitionSupport.js
```diff
@@ -0,0 +1,221 @@
...
+	* The the classname to apply for the "shown" (component is visible) resting state.
+	*
+	* @type {String}
+	* @default undefined
+	* @public
+	*/
+	shownClass: undefined,
+
+	/**
+	* The the classname to apply for the "hidden" (component is not visible) resting state.
```
aarontam: Minor thing: Double "the" here and for the following two properties.

src/ShowingTransitionSupport.js
```diff
@@ -0,0 +1,221 @@
...
+
+			// Prepare our visual state
+			this.applyStyle('display', null);
+			this.applyStyle('visibility', null);
+			if (this.showing) {
+				// Reset our state classes, in case we switched mid-stream
+				this.removeClass(this.hidingClass);
+				this.removeClass(this.hiddenClass);
+				sup.apply(this, args);
+				if (this.showingDuration && this.generated) {
```
aarontam: Wonder if we should check `this.hasNode()` instead.

src/ShowingTransitionSupport.js
```diff
@@ -0,0 +1,221 @@
...
+						this.removeClass(this.hidingClass);
+						this.addClass(this.hiddenClass);
+						this.set('showingTransitioning', false);
+						sup.apply(this, args);
+						this.applyStyle('visibility', 'hidden');
+						this.applyStyle('display', null);
+					}, this.hidingDuration);
+				} else {
+					// No transition, just a hidden class.
+					this.removeClass(this.hidingClass);
```
aarontam: This probably isn't necessary, right?

src/ShowingTransitionSupport.js
```diff
@@ -0,0 +1,221 @@
...
+				sup.apply(this, args);
+				if (this.showingDuration && this.generated) {
+					this.set('showingTransitioning', true);
+					// Start transition: Apply a class and start a timer.
+					// When timer finishes, run the exit function,
+					// remove the transitioning class
+					// and add the final-state class
+					this.addClass(this.showingClass);
+					this.startJob('showingTransition', function () {
+						utils.call(this, this.shownMethod);	// Run the supplied method.
```
aarontam: I wonder if we want to always assume that the scope of the `shownMethod` and the `hiddenMethod` is always on the instance itself vs. also on the owner, for example. This could come up if a component that mixes in `ShowingTransitionSupport` is declared in the components block of another control, and the methods are found on this control.

src/ShowingTransitionSupport.js
Djspaceg: It's necessary as back-up, in case the transition is cut short in the middle (for the edge case of removing the `hidingDuration` during the transition itself), it clears out the previous class so "hiding" and "hidden" don't exist on the node at the same time.

It's also very low cost. Reading a property and doing an string.index on it.

src/ShowingTransitionSupport.js
Djspaceg: So something like a new property that allows you to specify the scope, or something fancier? I assume you're meaning in the 2nd to last word "the `this` control" meaning the possessor of the components block; and not literally english "this", because english "this" would refer to the control that the mixin is mixed into, and we should not be putting functions on the components in the components block.

Anyway, what did you have in mind for making the `utils.call` method more versatile, instead of directly scoping it to `this`?

aarontam: Yep was referring to the possessor of the components block. Hadn't fully thought this through, but maybe for the case where the specified method is a string, we could first check the control that has mixed in `ShowingTransitionSupport`, and if the method doesn't exist in that scope, we can check the owner via `getInstanceOwner`. Adds a little bit more complexity, but I've seen this need come up throughout various use cases in the apps.

src/ShowingTransitionSupport.js
aarontam: Ah I didn't realize changing `hidingDuration` during the transition would cut the transition short. I agree that it is relatively low-cost, but seemed like it was covered by other cases (i.e. switching `showing` to `true` removes this class). Might also be a good safeguard to add a `stopJob` call for the hiding job when showing (and vice versa).

src/ShowingTransitionSupport.js
Djspaceg: Ah yes. Probably also a local var reference setup at create time so we don't need to check every time.

src/ShowingTransitionSupport.js
Djspaceg: Sorry. I see what you're getting at. I'll take that line out, and replace it and insert in the above `else` a stopJob call. Then if showing is toggled during a transition and duration is set to nothing, the job will be stopped. Also add some DurationChanged methods to do the same. Bulletproof.

src/ShowingTransitionSupport.js
Djspaceg: I believe I have something that does what you were requesting, among other changes. Strawman is also updated to include an example of this.

src/ShowingTransitionSupport.js
aarontam: Ack, my fault, @webOS101 says we need this for JSDoc purposes - I'll add this back in.

Djspaceg: I'd remembered needing this for a reason but forgot why. This is how the other mixins do it I believe.

aarontam: Funnily enough, in the process of looking into this, we found some mixins that are not documented correctly. :smile:

Djspaceg: Yuss! 🙏🏽

src/ShowingTransitionSupport.js
```diff
@@ -0,0 +1,242 @@
...
+				}
+			} else {
+				// Reset our state classes, in case we switched mid-stream
+				this.removeClass(this.showingClass);
+				this.removeClass(this.shownClass);
+				if (this.hidingDuration && this.hasNode()) {
+					this.set('showingTransitioning', true);
+					this.addClass(this.hidingClass);
+					this.startJob('showingTransition', function () {
+						utils.call(this._hiddenMethodScope, this.hiddenMethod);	// Run the supplied method.
```
ryanjduffy: I think the callback should fire when hiding/showing duration is 0/undefined as well. In cases where the duration would be configurable external to the control mixing ShowingTransitionSupport but the control relies on the callback to complete actions, it would need it in both cases. WDYT?

Djspaceg: That case does make sense. So also put the call line down on like 218 and 195?

ryanjduffy: Yup

Djspaceg: Done. (As well as a firing-order change.)

PR #1355, opened 2016-01-15
ENYO-2409: Fix Windows 10 app deployment error

germboy:
### Issue
Visual Studio outputs an error when packaging a Windows 10 enyo application.

### Fix
Validate that a method exists before firing, as Windows 10 removes the Windows 8 method and requirement of using `MSApp.execUnsafeLocalFunction` to set innerHTML content.

### DIFF
--- a/src/dom.js
+++ b/src/dom.js
@@ -576,7 +576,7 @@ var dom = module.exports = {
 };
 
 // override setInnerHtml for Windows 8 HTML applications
-if (typeof global.MSApp !== 'undefined') {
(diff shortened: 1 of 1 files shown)


src/dom.js
```diff
@@ -576,7 +576,7 @@ var dom = module.exports = {
 };
 
 // override setInnerHtml for Windows 8 HTML applications
-if (typeof global.MSApp !== 'undefined') {
+if (typeof global.MSApp !== 'undefined' && global.MSApp.execUnsafeLocalFunction !== 'undefined') {
```
aarontam: We are missing a `typeof` here. I saw this right when I clicked "merge" :disappointed: Going to revert this...

PR #1356, opened 2016-01-15
ENYO-2486 ENYO-3012 Add edge detection, fix docs

webOS101:

### DIFF
--- a/src/platform.js
+++ b/src/platform.js
@@ -52,7 +52,7 @@ exports = module.exports = {
 	* The name of the platform that was detected or `undefined` if the platform
 	* was unrecognized. This value is the key name for the major version of the
 	* platform on the exported object.
-	* @var {string} platformName
+	* @member {string} platformName
 	* @public
 	*/
 
--- a/src/platform.js
+++ b/src/platform.js
@@ -10,6 +10,7 @@ var utils = require('./utils');
 * * androidChrome (Chrome on Android, standard starting in 4.1)
 * * androidFirefox
 * * ie
+* * edge
 * * ios
 * * webos
 * * windowsPhone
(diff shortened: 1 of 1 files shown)


src/platform.js
```diff
@@ -35,17 +36,28 @@ var utils = require('./utils');
...
+	* `true` if the platform has native double-finger [events]{@glossary event}.
+	* @public
+	*/
 	gesture: Boolean(('ongesturestart' in window) || window.navigator.msMaxTouchPoints)
+
+	/**
+	* The name of the platform that was detected or `undefined` if the platform
+	* was unrecognized. This value is the key name for the major version of the
+	* platform on the exported object.
+	* @var {string} platformName
```
aarontam: What does `@var` mean in JSDoc?

webOS101: http://usejsdoc.org/tags-member.html  We're documenting a virtual member.  I suppose that @member will work as well... But I thought I tried it and settled on @var.  I can do a quick test again with @member to be sure.

aarontam: Gotcha - for some reason that synonym didn't come up in the search. I'm not opposed to it, but we seem to be using `@member` everywhere else.

src/platform.js
```diff
@@ -35,17 +36,28 @@ var utils = require('./utils');
...
+	* `true` if the platform has native double-finger [events]{@glossary event}.
+	* @public
+	*/
 	gesture: Boolean(('ongesturestart' in window) || window.navigator.msMaxTouchPoints)
+
+	/**
+	* The name of the platform that was detected or `undefined` if the platform
+	* was unrecognized. This value is the key name for the major version of the
+	* platform on the exported object.
+	* @member {string} platformName
```
aarontam: Oh missed this earlier, shouldn't this be `{String}`?

PR #1358, opened 2016-01-18
ENYO-3016: Fix touch-gesture platform identification for Windows 10, …

germboy:
…add Windows 10 Mobile

### Issue
enyo.platform reports the wrong touch/gesture information for Windows 10 and Windows 10 Mobile. Also Windows 10 Mobile is incorrectly identified as 'androidChrome'

### Fix
Enyo is using deprecated Windows pointer events (for IE 10 support - not to be confused with Windows 10) to determine touch/gesture capability. Since IE11, Microsoft has dropped support for its vendor-prefixed event "msMaxTouchPoints". I have now added support for "maxTouchPoints", while keeping the vendor-prefix version (until we decide we're completely dropping IE10 support).

I've also moved the Windows Phone platform regex match to the top - above 'androidChrome' - since Windows 10 Mobile contains a match to it, falsely being identified by enyo as 'androidChrome'.

### DIFF
--- a/src/platform.js
+++ b/src/platform.js
@@ -46,7 +46,7 @@ exports = module.exports = {
 	* `true` if the platform has native double-finger [events]{@glossary event}.
 	* @public
 	*/
-	gesture: Boolean(('ongesturestart' in window) || window.navigator.msMaxTouchPoints || window.navigator.maxTouchPoints)
+	gesture: Boolean(('ongesturestart' in window) || (window.navigator.msMaxTouchPoints && window.navigator.msMaxTouchPoints > 1) || (window.navigator.maxTouchPoints && window.navigator.maxTouchPoints > 1))
 
 	/**
 	* The name of the platform that was detected or `undefined` if the platform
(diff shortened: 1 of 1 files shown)


src/platform.js
```diff
@@ -41,12 +41,12 @@ exports = module.exports = {
 	* `true` if the platform has native single-finger [events]{@glossary event}.
 	* @public
 	*/
-	touch: Boolean(('ontouchstart' in window) || window.navigator.msMaxTouchPoints),
+	touch: Boolean(('ontouchstart' in window) || window.navigator.msMaxTouchPoints || window.navigator.maxTouchPoints),
```
aarontam: Should we actually be checking if `msMaxTouchPoints`/`maxTouchPoints` is exactly 1 here, and >1 for the gesture feature detection below?

germboy: I agree that we should probably be checking for >1 below for gesture events, but we shouldn't be checking for *exactly* 1 here - as we need to check for the existence of at least 1 touch point. It's likely that maxTouchPoints will return more than 1.

aarontam: Ah yes, my mistake - somehow was considering these to be mutually exclusive. I think the existing check is good, and we should change the other check to `> 1`.

PR #1362, opened 2016-01-23
PLAT-15150: Fix issues with destruction.

aarontam:
### Issue
There were a few issues when disabling `cacheViews`, namely related to the panel lifecycle methods that assumed that certain components were still instantiated (primarily in the `moonstone/LightPanels` case), which caused null reference errors as the panel had already been destroyed.

### Fix
There were a number of tweaks that had to be made to support destruction of panels, as this was not a primary use case and the related logic for this mode had gotten out of sync.
- The post-transition logic has been adjusted:
  - Instead of only executing the `postTransition` method of a panel asynchronously, we are now asynchronously executing the entirety of the post-transition work, including `postTransition` and panel removal. Unfortunately, there is still a performance issue if `postTransition` is not made asynchronous, and as a result, we cannot destroy the panel (in a non-caching mode) before `postTransition` is executed. This also necessitated making a tweak to how we are replacing panels, as a similar logical fallacy would occur if we did not defer the actual destruction of the replaced panels.
  - Deferred purging of panels has been re-introduced. This was originally a speculative fix for some hard-to-reproduce issues that still appeared after this fix went in, and were seemingly related to `enyo/History` + the back key, instead. This allows us to purge, and thus destroy, panels after the lifecycle methods have been executed.
- The `popOnBack` property now takes precedence over `cacheViews`. We wanted to support the case of destroying panels we were returning from, while caching panels we had visited when moving forward (i.e. destroying panels we likely would not revisit, and caching panels we would have a high-likelihood of revisiting). As a result, the default value of this property has been changed to `false` to preserve existing behavior.
- We have introduced a `popView` method to `enyo/ViewPreloadSupport`, that can be called directly and is also used by `restoreView`. This allows for clean-up of views (and references to them) that had been previously cached and were explicitly destroyed, such as when `popOnBack` is enabled.
- The `removePanels` method has been correctly marked as `private` as it is not useful for public consumption, due to intrinsic knowledge of whether to destroy or cache a panel, based on the navigation direction.

### DIFF
--- a/src/LightPanels/LightPanels.js
+++ b/src/LightPanels/LightPanels.js
@@ -488,11 +488,11 @@ module.exports = kind(
 
 		if (direction < 0) {
 			for (i = panels.length - 1; i > index; i--) {
-				this.removePanel(panels[i]);
+				this.removePanel(panels[i], this.popOnBack);
 			}
 		} else {
 			for (i = 0; i < index; i++) {
-				this.removePanel(panels[i], true);
+				this.removePanel(panels[i], !this.cacheViews, true);
 			}
 		}
 	},
@@ -501,17 +501,19 @@ module.exports = kind(
 	* Removes the specified panel.
 	*
 	* @param {Object} panel - The panel to remove.
+	* @param {Boolean} destroy - Whether or not the panel should be destroyed; if `false`, the panel
+	*	will instead be cached.
 	* @param {Boolean} [preserve] - If {@link module:enyo/LightPanels~LightPanels#cacheViews} is
 	*	`true`, this value is used to determine whether or not to preserve the current panel's
 	*	position in the component hierarchy and on the screen, when caching.
 	* @private
 	*/
-	removePanel: function (panel, preserve) {
+	removePanel: function (panel, destroy, preserve) {
 		if (panel) {
-			if (this.cacheViews) {
-				this.cacheView(panel, preserve);
-			} else {
+			if (destroy) {
 				panel.destroy();
+			} else {
+				this.cacheView(panel, preserve);
 			}
 		}
 	},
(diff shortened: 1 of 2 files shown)


src/LightPanels/LightPanels.js
```diff
@@ -530,17 +485,16 @@ module.exports = kind(
...
 		end = start + count;
 		insertBefore = panels[end];
 		commonInfo = {addBefore: insertBefore};
+		queue = queue || [];
 
-		// remove existing panels
-		for (idx = end - 1; idx >= start; idx--) {
-			this.removePanel(panels[idx]);
-		}
+		queue = queue.concat(panels.splice(start, end - start));
```
webOS101: Does queue serve a purpose if we don't use it?

webOS101: Also, is it OK that we don't call removePanel() anymore?  This won't call panel.destroy(), for one thing.

src/ViewPreloadSupport.js
```diff
@@ -142,13 +142,27 @@ var ViewPreloadSupport = {
...
+		}
+
+		return view;
+	},
+
+	/**
+	* Pops the specified view that was previously cached.
+	*
+	* @param {String} id - The unique identifier for the cached view that is being popped.
+	* @public
```
webOS101: Let's go ahead and doc the return value

src/LightPanels/LightPanels.js
aarontam: The intention was that we would modify `this.popQueue`, which we have a reference to if it exists, but this won't work for the case where we're creating a new array, so I'll fix this up. `panel.destroy()` will be called when we actually process our queue of panels to pop, if we are not caching views.

src/ViewPreloadSupport.js
aarontam: Aye!

PR #1367, opened 2016-01-28
ENYO-3030: Prevent reading disabled component

kbs12e:
## Issue
screen reader doesn't read disable Button's string, but it reads disabled IconButton.

## Cause
Button is created with button tag, but IconButton is div tag. If form element has disabled attribute,
screen reader doesn't read string.

## Fix
Prevent reading disabled component's string as removing tabindex.

https://jira2.lgsvl.com/browse/ENYO-3030
Enyo-DCO-1.1-Signed-off-by: Bongsub Kim <bongsub.kim@lgepartner.com>


seunghoh:
Close as outdated PR

PR #1373, opened 2016-02-05
ENYO-3080: requestAnimationFrame not in usage for iOS >=7

JayCanuck:
Enyo-DCO-1.1-Signed-off-by: Jason Robitaille <jason.robitaille@lge.com>

### DIFF
--- a/src/animation.js
+++ b/src/animation.js
@@ -33,8 +33,8 @@ var _cancelFrame = function(inId) {
 };
 
 for (var i = 0, pl = prefix.length, p, wc, wr; (p = prefix[i]) || i < pl; i++) {
-	// if we're on ios 6 just use setTimeout, requestAnimationFrame has some kinks currently
-	if (platform.ios >= 6) {
+	// if we're on ios 6 just use setTimeout, requestAnimationFrame has some kinks
+	if (platform.ios < 7) {
 		break;
 	}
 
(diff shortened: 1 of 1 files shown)


src/animation.js
```diff
@@ -33,8 +33,8 @@ var _cancelFrame = function(inId) {
 };
 
 for (var i = 0, pl = prefix.length, p, wc, wr; (p = prefix[i]) || i < pl; i++) {
-	// if we're on ios 6 just use setTimeout, requestAnimationFrame has some kinks currently
-	if (platform.ios >= 6) {
+	// if we're on ios 6 just use setTimeout, requestAnimationFrame has some kinks
+	if (platform.ios < 7) {
```
aarontam: Interesting - for iOS 5 and below, it previously fell back to `setTimeout`, as well, because the conditional on [L45](https://github.com/enyojs/enyo/pull/1373/files#diff-2314e4f2e74b0a130fa863211c157653L45) failed. Wonder if we should let these iOS versions run through this feature detection logic, in case something changes in the future, and only make a special case for iOS 6. Otherwise, we should update the comment to be clearer about our intention here.

JayCanuck: You mean for iOS5 and below?

aarontam: No, I mean exclusively for iOS 6. :smile: It doesn't change the ultimate result and these are old platform versions anyways, but I wanted to preserve the intent of the feature detection here, rather than start a trend of having blanket conditionals. iOS 6 is the exception case we need to explicitly handle because it actually has an implementation of `requestAnimationFrame` (and would pass our feature detection), but that implementation is quirky/buggy and does not work properly, which is why we added the conditional in the first place.

PR #1382, opened 2016-02-12
Fixes ENYO-3033

ryanjduffy:
In order to eliminate string references to kindNames and avoid circular
dependencies between the private kinds used by RelationalModel, some
refactoring was required including some post-kind setup in the module's
index.js


src/RelationalModel/toOne.js
```diff
@@ -65,9 +69,6 @@ var toOne = module.exports = kind(
 		// ensure we have a valid model constructor
 		if (typeof model == 'string') model = kind.constructorForKind(model);
```
aarontam: Can we remove this, as well?

src/RelationalModel/Relation.js
```diff
@@ -107,7 +115,7 @@ var Relation = module.exports = kind(
 	/**
 	* @private
 	*/
-	setRelated: function (related) {
+	setRelated: function (related, opts) {
```
aarontam: Curious if this having this unused parameter here was intentional, to match the overriding done by its subkinds.

ryanjduffy: It was intentional. Adding a note to that effect.

src/RelationalModel/Relation.js
```diff
@@ -100,14 +101,25 @@ var Relation = module.exports = kind(
...
+	*/
 	getRelated: function () {
 		return this.related;
 	},
 
 	/**
+	* Sets the `related` model
+	*
+	* @param {[type]} related - The related model
+	* @param {[type]} [opts] - Used by subkinds to configure the behavior of the method
```
webOS101: I think related should probably be {module:enyo/Model~Model}  and opts should be {Object} or, if we have a typedef form options, that.

src/RelationalModel/toMany.js
```diff
@@ -130,15 +129,22 @@ var toMany = module.exports = kind(
...
+	*/
+	isRelated: function (related) {
+		return this.related.indexOf(related) >= 0;
+	},
 	
 	/**
 	* @private
 	*/
-	setRelated: function (data) {
+	setRelated: function (data, opts) {
```
webOS101: Might as well comment this one, too!

src/RelationalModel/Relation.js
ryanjduffy: oops ...

PR #1384, opened 2016-02-16
ENYO-3011: Ensure that transition clean-up logic is always executed.

aarontam:
### Issue
When a transitioning element is hidden, the `ontransitionend` event is never fired (https://bugs.webkit.org/show_bug.cgi?id=51945). In the case of `LightPanels`, this prevents some necessary post-transition clean-up that is executed as a result of handling this event.

### Fix
We have added some fallback logic that directly executes the clean-up logic, if it has not already been executed for the current transition. An alternative, and arguably more direct solution, could be to execute the clean-up logic whenever the panels instance is hidden (https://github.com/enyojs/enyo/compare/2.6.0-dev...ENYO-3011-aarontam-1). This works well for the current case and the specific behavior of hiding transitioning elements, but it seemed safer to have a catch-all to handle any other edge cases and quirks we may not be aware of.

### DIFF
--- a/src/LightPanels/LightPanels.js
+++ b/src/LightPanels/LightPanels.js
@@ -677,13 +677,17 @@ module.exports = kind(
 	*
 	* @param {Object} sender - The event sender.
 	* @param {Object} ev - The event object.
-	* @param {Boolean} [direct] - If `true`, this was a non-animated (direct) transition.
 	* @private
 	*/
-	transitionFinished: function (sender, ev, direct) {
+	transitionFinished: function (sender, ev) {
 		var prevPanel, currPanel;
 
-		if (this.transitioning && ((ev && ev.originator === this.$.client) || direct)) {
+		if (this.transitioning && (!ev || ev.originator === this.$.client)) {
+			if (this._fallbackTimeout) {
+				global.clearTimeout(this._fallbackTimeout);
+				this._fallbackTimeout = null;
+			}
+
 			prevPanel = this._previousPanel;
 			currPanel = this._currentPanel;
 
@@ -805,7 +809,8 @@ module.exports = kind(
 		this._currentPanel = nextPanel;
 
 		// ensure that `transitionFinished` is called in the case where we are not animating
-		if (!this.shouldAnimate() || !animate) this.transitionFinished(null, null, true);
+		if (!this.shouldAnimate() || !animate) this.transitionFinished();
+		else this.setupFallback();
 	},
 
 	/**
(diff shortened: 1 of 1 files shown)


src/LightPanels/LightPanels.js
```diff
@@ -862,6 +867,13 @@ module.exports = kind(
 		for (var idx = 0; idx < viewProps.length; idx++) {
 			this.removeTask(this.getViewId(viewProps[idx]));
 		}
+	},
+
+	/**
+	* @private
+	*/
+	setupFallback: function () {
+		this._fallbackTimeout = global.setTimeout(this.bindSafely('transitionFinished'), this.duration + 100);
```
ryanjduffy: Wondering if the extra 100ms is necessary here. Do we care if the fallback fires a moment before the `ontransitionend` event? Is there a risk the transition might not be done in `this.duration` ms but will be in `this.duration + 100`ms and that the user would notice?

src/LightPanels/LightPanels.js
```diff
@@ -805,7 +809,8 @@ module.exports = kind(
 		this._currentPanel = nextPanel;
 
 		// ensure that `transitionFinished` is called in the case where we are not animating
-		if (!this.shouldAnimate() || !animate) this.transitionFinished(null, null, true);
+		if (!this.shouldAnimate() || !animate) this.transitionFinished();
```
ryanjduffy: This is more a stylistic question but I've been wondering if we should get out of the habit of calling event handlers directly and instead refactor the logic into another private method. It's minor in this case but allows us to avoid conditional arg requests (`(!ev || ev.originator === this.$.client)`). We'd still need similar logic in the handler but the shared method would just guard on `this.transitioning`.

wdyt?

aarontam: Good question. :smile: I had actually wavered on this and split this into a `cleanUpTransition` method and a `transitionFinished` handler as it *looked* cleaner. I ultimately scrapped this in favor of the current solution, to not add another method, and to reduce the surface area of this change. If this is a pattern we want to follow, I am definitely not opposed to making that change, but maybe it makes more sense in a next-gen version?

src/LightPanels/LightPanels.js
aarontam: It probably isn't absolutely necessary, but I wanted to have some safeguard so that the clean-up logic is not executed too early. Some of this logic hides the transitioning-out panel, so there could be a minor visual blip. Also, we are effectively using a "mask" to cover the transitioning panels, to prevent user clicks, and this is removed in our clean-up logic, so there is a (minor, though not entirely implausible due to QA practices) risk that something gets triggered/activated during transition by the user. I will say that `100ms` is slightly arbitrary, though. I'm open to something more definitive, or if you feel that the risk of these occurrences (and the potential impact) is low enough to do away with the offset.

PR #1387, opened 2016-02-19
ENYO-3019: IE8 Support Removal

germboy:
Removing core support for IE8

### DIFF
--- a/src/Ajax.js
+++ b/src/Ajax.js
@@ -184,7 +184,7 @@ var Ajax = module.exports = kind(
 		}
 		utils.mixin(xhr_headers, this.headers);
 		// don't pass in headers structure if there are no headers defined as this messes
-		// up CORS code for IE8-9
+		// up CORS code for IE9
 		if (utils.keys(xhr_headers).length === 0) {
 			xhr_headers = undefined;
 		}
@@ -278,7 +278,7 @@ var Ajax = module.exports = kind(
 				}
 			} else {
 				var text = '';
-				// work around IE8-9 bug where accessing responseText will thrown error
+				// work around IE9 bug where accessing responseText will thrown error
 				// for binary requests.
 				if (typeof inXhr.responseText === 'string') {
 					text = inXhr.responseText;
@@ -337,19 +337,15 @@ var Ajax = module.exports = kind(
 	* @private
 	*/
 	updateProgress: function (event) {
-		// IE8 doesn't properly support progress events and doesn't pass an object to the
-		// handlers so we'll check that before continuing.
-		if (event) {
-			// filter out 'input' as it causes exceptions on some Firefox versions
-			// due to unimplemented internal APIs
-			var ev = {};
-			for (var k in event) {
-				if (k !== 'input') {
-					ev[k] = event[k];
-				}
+		// filter out 'input' as it causes exceptions on some Firefox versions
+		// due to unimplemented internal APIs
+		var ev = {};
+		for (var k in event) {
+			if (k !== 'input') {
+				ev[k] = event[k];
 			}
-			this.sendProgress(event.loaded, 0, event.total, ev);
 		}
+		this.sendProgress(event.loaded, 0, event.total, ev);
 	},
 	
 	/**
--- a/src/dispatcher.js
+++ b/src/dispatcher.js
@@ -8,8 +8,7 @@ require('enyo');
 var
 	logger = require('./logger'),
 	master = require('./master'),
-	utils = require('./utils'),
-	platform = require('./platform');
+	utils = require('./utils');
 
 var
 	Dom = require('./dom');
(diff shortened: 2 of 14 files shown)


src/Ajax.js
```diff
@@ -337,19 +337,15 @@ var Ajax = module.exports = kind(
...
+		// filter out 'input' as it causes exceptions on some Firefox versions
+		// due to unimplemented internal APIs
+		var ev = {};
+		for (var k in event) {
+			if (k !== 'input') {
+				ev[k] = event[k];
 			}
-			this.sendProgress(event.loaded, 0, event.total, ev);
 		}
+		this.sendProgress(event.loaded, 0, event.total, ev);
```
webOS101: Did you mean to move this outside the loop?  Was it a mistake to have inside the loop to begin with?

src/utils.js
```diff
@@ -1026,15 +1023,13 @@ Array.prototype.indexOf = Array.prototype.indexOf || function (el, offset) {
...
-* advise, moving forward, that the built-in method be used. But to preserve our original API, it
-* will simply call this method, knowing it exists.
+* Because our older API parameters are not consistent with other array API methods, we advise,
+* moving forward, that the built-in method be used. But to preserve our original API, it will
+* simply call this method, knowing it exists.
 *
 * @private
 */
-Array.prototype.lastIndexOf = Array.prototype.lastIndexOf || function (el, offset) {
+Array.prototype.lastIndexOf = function (el, offset) {
```
webOS101: This is the wrong approach.  If we need only for IE8 then this whole thing can be removed.  Same for indexOf.  This overwrites the built-in with ours, which we absolutely don't want.

src/Ajax.js
germboy: It was always outside the loop. I simply removed the `if (event)` conditional for IE8 and shifted the tabs to the left. The way the github changes are laid out, gives the appearance that it moved, when it actually didn't.

webOS101: aha!  Brilliant!

src/dispatcher.js
```diff
@@ -86,22 +85,9 @@ var dispatcher = module.exports = dispatcher = {
...
-				inListener.attachEvent("on" + inEvent, function(e) {
-					e.target = e.srcElement;
-					if (!e.preventDefault) {
-						e.preventDefault = this.iePreventDefault;
-					}
-					return (inHandler || dispatch)(e);
-				});
-			};
-		}
+		this.listen = function(inListener, inEventName, inHandler) {
```
ryanjduffy: There's no need for the reassignment logic any longer since we're not altering based on platform. We can just update the body of `listen()` to be
```
listen: function (inListener, inEventName, inHandler) {
	inListener.addEventListener(inEventName, inHandler || dispatch, false);
}
```

src/dispatcher.js
```diff
@@ -109,16 +95,9 @@ var dispatcher = module.exports = dispatcher = {
...
-			this.stopListening = function(inListener, inEventName, inHandler) {
-				inListener.removeEventListener(inEventName, inHandler || dispatch, false);
-			};
-		} else {
-			//enyo.log("IE8 COMPAT: using 'detachEvent'");
-			this.stopListening = function(inListener, inEvent, inHandler) {
-				inListener.detachEvent("on" + inEvent, inHandler || dispatch);
-			};
-		}
+		this.stopListening = function(inListener, inEventName, inHandler) {
```
ryanjduffy: Same as above; no need for the reassignment.

src/dom.js
```diff
@@ -216,64 +186,11 @@ var dom = module.exports = {
 	/**
 	* @private
 	*/
-	// Workaround for lack of compareDocumentPosition support in IE8
-	// Code MIT Licensed, John Resig; source: http://ejohn.org/blog/comparing-document-position/
-	compareDocumentPosition: function(a, b) {
```
ryanjduffy: Spotlight uses `dom.compareDocumentPosition()` so we'll need to either keep a facade here or update spotlight.

https://github.com/enyojs/spotlight/blob/2.6.0-dev/src/container.js#L278

src/dom.js
```diff
@@ -377,11 +293,7 @@ var dom = module.exports = {
 				}
 				// Need to correct for borders if any exist on parent elements
 				if (node !== targetNode) {
-					if (node.currentStyle) {
-						// Oh IE, we do so love working around your incompatibilities
-						borderLeft = parseInt(node.currentStyle.borderLeftWidth, 10);
-						borderTop = parseInt(node.currentStyle.borderTopWidth, 10);
-					} else if (global.getComputedStyle) {
+					if (global.getComputedStyle) {
```
ryanjduffy: Don't think we need this guard any longer though it's no clear why we needed it before with the existing IE guard above.

http://caniuse.com/#search=getcomputedstyle

src/ready.js
```diff
@@ -67,18 +67,14 @@ init = function (event) {
 * @private
 */
 add = function (event, fn) {
-	var name = doc.addEventListener? "addEventListener": "attachEvent";
-	var on = name === "attachEvent"? "on": "";
-	doc[name](on + event, fn, false);
+	doc["addEventListener"](event, fn, false);
```
ryanjduffy: No need for the stringified method name.

src/ready.js
```diff
@@ -67,18 +67,14 @@ init = function (event) {
...
 };
 
 /**
 * @private
 */
 remove = function (event, fn) {
-	var name = doc.addEventListener? "removeEventListener": "detachEvent";
-	var on = name === "detachEvent"? "on": "";
-	doc[name](on + event, fn, false);
+	doc["removeEventListener"](event, fn, false);
```
ryanjduffy: Same as above

ryanjduffy:
Based on MDN's support matrices, most of the other Array polyfills (except `find` and `findIndex`) can be removed as well.

ryanjduffy:
I'd like to verify if the conditional in `Select.render()` is still required. Couldn't tell from the referenced MS support article.

src/dom.js
germboy: Ah, good call. I've updated spotlight in a related PR.

src/dom.js
germboy: After removing the IE8-only `currentStye` support, there are conditions to use `getComputedStyle` or a node's `style` prop value directly. I wasn't too sure about that what other platforms we support that *don't* support `getComputedStyle` either (perhaps older webOS platform?). But since I'm dealing with IE8-only, I'd rather just remove the IE8 support and leave anything else. I'd suggest removing the check in another PR, once we realize what it's there for.

germboy:
@ryanjduffy The conditional in `Select.render()` is still required for now, as IE's select-innerHTML bug exists in not only IE8, but IE9 also. We could probably update the conditional at some point to reflect that: `if (platform.ie < 10) {`

src/dom.js
ryanjduffy: That works. Here's the originating PR for reference:

https://github.com/enyojs/enyo/pull/209

src/dispatcher.js
```diff
@@ -220,14 +193,12 @@ function dispatch (inEvent) {
...
 dispatcher.bubble = function(inEvent) {
-	// '|| window.event' clause needed for IE8
-	var e = inEvent || global.event;
-	if (e) {
-		// We depend on e.target existing for event tracking and dispatching.
-		if (!e.target) {
-			e.target = e.srcElement;
+	if (inEvent) {
+		// We depend on inEvent.target existing for event tracking and dispatching.
+		if (!inEvent.target) {
```
aarontam: I think we can remove this entirely as it was IE-specific and IE9+ supports `Event.target`.

PR #1392, opened 2016-02-24
Fixes ENYO-3126 add docs for RepeaterChildSupport#selectedClass

ryanjduffy:

### DIFF
--- a/src/RepeaterChildSupport.js
+++ b/src/RepeaterChildSupport.js
@@ -40,6 +40,15 @@ var RepeaterChildSupport = {
 	selected: false,
 
 	/**
+	* The CSS class applied to the child when selected
+	*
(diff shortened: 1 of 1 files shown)


src/RepeaterChildSupport.js
```diff
@@ -40,6 +40,14 @@ var RepeaterChildSupport = {
 	selected: false,
 
 	/**
+	* The CSS class applied to the child when selected
+	*
+	* @type {String}
+	* @default null
+	* @public
+	*/
```
aarontam: Did you forgot to specify the property and/or the `@name` pragma? :smile:

ryanjduffy: I pulled it because it was overwriting the value I added in onyx. Adding the `@name` in its place.

PR #1397, opened 2016-02-26
ENYO-3130: Touch/gesture fix

germboy:
### Issue
Certain platforms will return an incorrect platform touch/gesture value, due to a recent change. 

### Fix
As this issue was introduced after adding specific IE-Edge browser support, we need to verify that only Edge/IE are affected by the new conditional(s) to check touch/gesture support. IE9 does not support touch, so we don't worry about that.

Gesture Support
IE10, IE11, and Edge all support the proprietary `onmsgesturestart` event, so we can simply check for this to determine MS gesture support.

Touch Support
IE10 supports the already-in-use `msMaxTouchPoints` property and Edge/IE11 support `msManipulationViewsEnabled` and `maxTouchPoints`, both which return the touch manipulation ability and if touch points are available.

### DIFF
--- a/src/platform.js
+++ b/src/platform.js
@@ -41,13 +41,12 @@ exports = module.exports = {
 	* `true` if the platform has native single-finger [events]{@glossary event}.
 	* @public
 	*/
-	touch: Boolean(('ontouchstart' in window) || window.navigator.msMaxTouchPoints || window.navigator.maxTouchPoints),
+	touch: Boolean(('ontouchstart' in window) || window.navigator.msMaxTouchPoints || (window.navigator.msManipulationViewsEnabled && window.navigator.maxTouchPoints)),
 	/**
 	* `true` if the platform has native double-finger [events]{@glossary event}.
 	* @public
 	*/
-	gesture: Boolean(('ongesturestart' in window) || (window.navigator.msMaxTouchPoints && window.navigator.msMaxTouchPoints > 1) || (window.navigator.maxTouchPoints && window.navigator.maxTouchPoints > 1))
-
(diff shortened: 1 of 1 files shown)


src/platform.js
```diff
@@ -41,12 +41,12 @@ exports = module.exports = {
...
 	* @public
 	*/
-	touch: Boolean(('ontouchstart' in window) || window.navigator.msMaxTouchPoints || window.navigator.maxTouchPoints),
+	touch: Boolean(('ontouchstart' in window) || window.navigator.msMaxTouchPoints || (window.navigator.msManipulationViewsEnabled && window.navigator.maxTouchPoints)),
 	/**
 	* `true` if the platform has native double-finger [events]{@glossary event}.
 	* @public
 	*/
-	gesture: Boolean(('ongesturestart' in window) || (window.navigator.msMaxTouchPoints && window.navigator.msMaxTouchPoints > 1) || (window.navigator.maxTouchPoints && window.navigator.maxTouchPoints > 1))
+	gesture: Boolean(('ongesturestart' in window) || ('onmsgesturestart' in window && window.navigator.msMaxTouchPoints && window.navigator.msMaxTouchPoints > 1) || ('onmsgesturestart' in window && window.navigator.maxTouchPoints && window.navigator.maxTouchPoints > 1))
```
aarontam: I realize this existed from the previous change, but I wonder if we can make this conditional more compact and take out `window.navigator.msMaxTouchPoints` in the second part of the conditional, and also `window.navigator.maxTouchPoints` in the third part of the conditional, as the `>1` comparison will return `false` if these properties are undefined. I think this still expresses the same intention - what do you think? Or maybe wrap the conditional across a few lines.

Another alternative:
```
gesture: Boolean(('ongesturestart' in window) || ('onmsgesturestart' in window && (window.navigator.msMaxTouchPoints > 1 || window.navigator.maxTouchPoints > 1)))
```

PR #1399, opened 2016-02-26
PLAT-16501 moon.Input provides validator to prevent inputting invalid number

Sangwook-Lee:
### Issue
This is new UX requirement in 2017 TV. 
User can input the invalid value out of correct range. So moon.input provides validator to prevent inputting invalid number.

### Fix
I added published variables(min, max) to make range from default values.
Also I implemented change handler to make it dynamically.
If moon.input has number type, min and max value, contexturePopup is shown by validity when user input the wrong number.

Enyo-DCO-1.1-Signed-off-by: Sangwook Lee <sangwook1203.lee@lge.com>

### DIFF
--- a/src/Input/Input.js
+++ b/src/Input/Input.js
@@ -131,14 +131,34 @@ module.exports = kind(
 		* @default false
 		* @public
 		*/
-		selectOnFocus: false
+		selectOnFocus: false,
+
+		/**
+		* The min attribute specifies the minimum value for an [input]{@link module:enyo/Input~Input}.
+		*
+		* @type {Number}
+		* @default null
+		* public
+		*/
+		min: null,
+
+		/**
+		* The max attribute specifies the maximum value for an [input]{@link module:enyo/Input~Input}.
+		*
+		* @type {Number}
+		* @default null
+		* public
+		*/
+		max: null
 	},
 
 	/**
 	* @private
 	*/
 	events: {
-		onDisabledChange: ''
+		onDisabledChange: '',
+		onenabledValidity: '',
+		ondisabledValidity: ''
 	},
 
 	/**
@@ -149,6 +169,11 @@ module.exports = kind(
 	* @public
 	*/
 	defaultFocus: false,
+
+	/**
+	* @private
+	*/
+	_enabledValidity: false,
 
 	/**
 	* @private
@@ -212,6 +237,26 @@ module.exports = kind(
 			}
 		};
 	}),
+
+	/**
+	* @private
+	*/
+	minChanged: function () {
+		this.setAttribute('min', this.min);
+		if(this.type == 'number') {
+			this.set('_enabledValidity', true);
+		}
+	},
+
+	/**
+	* @private
+	*/
+	maxChanged: function () {
+		this.setAttribute('max', this.max);
+		if(this.type == 'number') {
+			this.set('_enabledValidity', true);
+		}
+	},
 
 	/**
 	* @private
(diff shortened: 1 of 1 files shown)


src/Input/Input.js
```diff
@@ -323,9 +368,16 @@ module.exports = kind(
 	/**
 	* @private
 	*/
-	input: function () {
+	input: function (sender, e) {
 		var val = this.getNodeProperty('value');
 		this.setValue(val);
+		if(this._enabledValidity) {
+			if(!e.target.validity.valid) {
```
otisnate: You can distinguish each case using ```validity.rangeUnderFlow``` and ```validity.rangeOverflow``` when ```min``` and ```max``` property have value

src/Input/Input.js
```diff
@@ -131,14 +131,34 @@ module.exports = kind(
...
+		max: null
 	},
 
 	/**
 	* @private
 	*/
 	events: {
-		onDisabledChange: ''
+		onDisabledChange: '',
+		onenabledValidity: '',
```
KunmyonChoi: The 'onEnabledValidity' would be better to custom event name.

aarontam: What about `onValidityEnabled`?

src/Input/Input.js
```diff
@@ -323,9 +368,16 @@ module.exports = kind(
 	/**
 	* @private
 	*/
-	input: function () {
+	input: function (sender, e) {
 		var val = this.getNodeProperty('value');
 		this.setValue(val);
+		if(this._enabledValidity) {
+			if(!e.target.validity.valid) {
+				this.bubble('onenabledValidity');
```
aarontam: Let's use the `this.do[eventName]` convention i.e. `this.doValidityEnabled()`.

KunmyonChoi:
- I would like to have a property of validityCheck to enable or disable this feature. Do validity check only when the value is true and type is number.
- The code is in both of Enyo input and Moonstone inputDecorator. But, the code looks like can not use this feature without inputDecorator. So, if you want moonstone specific implementation then it would be better in moonstone only. If you want to move some of code to enyo then the code should not know about moonstone.

src/Input/Input.js
```diff
@@ -216,6 +241,26 @@ module.exports = kind(
 	/**
 	* @private
 	*/
+	minChanged: function () {
+		this.setAttribute('min', this.min);
+		if(this.type == 'number') {
```
KunmyonChoi: It looks like _enableValidity variable doesn't seem to do much here.

src/Input/Input.js
```diff
@@ -323,9 +368,16 @@ module.exports = kind(
 	/**
 	* @private
 	*/
-	input: function () {
+	input: function (sender, e) {
 		var val = this.getNodeProperty('value');
 		this.setValue(val);
+		if(this._enabledValidity) {
```
KunmyonChoi: I would suggest to make enableValidity as published property and use here. So, user can explicitly set it as true when he want to enable validity check feature.

src/Input/Input.js
KunmyonChoi: You can shorten this as single line.
this.bubble(validity ? 'onenabledValidity' : 'ondisabledValidity');

src/Input/Input.js
```diff
@@ -216,6 +241,26 @@ module.exports = kind(
...
+			this.set('_enabledValidity', true);
+		}
+	},
+
+	/**
+	* @private
+	*/
+	maxChanged: function () {
+		this.setAttribute('max', this.max);
+		if(this.type == 'number') {
```
KunmyonChoi: Please add space before (.

src/Input/Input.js
```diff
@@ -323,9 +368,16 @@ module.exports = kind(
...
 	* @private
 	*/
-	input: function () {
+	input: function (sender, e) {
 		var val = this.getNodeProperty('value');
 		this.setValue(val);
+		if(this._enabledValidity) {
+			if(!e.target.validity.valid) {
+				this.bubble('onenabledValidity');
+			}else {
```
KunmyonChoi: Please add space before and after 'else'.

PR #1402, opened 2016-03-02
ENYO-3158: Adjust Buffer size based on item size and count

KunmyonChoi:
### Issue:
We use page concept on DataList and DataGridList. When item count is not devided by controls per page, last page can have different page height from others. The adjustBuffer function of VerticalDelegate and VerticalGridDelegate calculate buffer size by accumulating page size. But, we use default page size when the page is not generated yet. So, the buffer size is different between before and after generating the last page.

### Fix:
Calculate buffer size based on item size and item collection length instead of page size and page height. And divide item collection length by number of columns in case of GridList.

Enyo-DCO-1.1-Signed-off-by: Kunmyon Choi kunmyon.choi@lge.com

### DIFF
--- a/src/VerticalGridDelegate.js
+++ b/src/VerticalGridDelegate.js
@@ -18,7 +18,7 @@ var
 var p = utils.clone(VerticalDelegate);
 kind.extendMethods(p, {
 	/** @lends enyo/VerticalGridDelegate */
-	
+
 	/**
 	* Once the list is initially rendered, it will generate its [scroller]{@link module:enyo/Scroller~Scroller}
 	* (so we know that is available). Now we need to cache our initial size values and apply
@@ -38,7 +38,7 @@ kind.extendMethods(p, {
 		// want to do any more initialization
 		if (list.collection && list.collection.length) { this.reset(list); }
 	},
-	
+
 	/**
 	* Resets the page, setting `canAddResetClass` flag if appropriate.
 	*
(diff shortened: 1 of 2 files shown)


src/VerticalGridDelegate.js
```diff
@@ -262,23 +262,19 @@ kind.extendMethods(p, {
...
 			bs = 0, sp = list.psizeProp, ss = list.ssizeProp,
-			n = list.$.buffer.node || list.$.buffer.hasNode(), p;
+			n = list.$.buffer.node || list.$.buffer.hasNode();
 		if (n) {
-			for (var i=0; i<pc; ++i) {
-				p = list.metrics.pages[i];
-				bs += (p && p[sp]) || ds;
-			}
-			bs += list.spacing;
+			bs = (Math.floor(count/list.columns)+(count%list.columns? 1: 0))*cs;
```
aarontam: Can we simplify this by using `Math.ceil`:

```
bs = Math.ceil(count / list.columns) * cs;
```

KunmyonChoi: To be honest, I don't really understand what (count%list.columns? 1: 0) doing in this calculation. I borrowed this line from pageHeight function to calculate buffer size. So, if you say that then I think we could think simplify pageHeight also. :)

aarontam: Ah ha, yeah let's change it if you don't see any issues with using `Math.ceil` instead.

aarontam:
Making a note that we'll need to deal with Travis and the whole NPM module debacle, separately, as this branch passed a few days ago, and all that was done was a rebase just now.

PR #1405, opened 2016-03-07
ENYO-3175: Added support for Mozilla based browsers

Djspaceg:

### DIFF
--- a/src/SpriteAnimation/SpriteAnimation.js
+++ b/src/SpriteAnimation/SpriteAnimation.js
@@ -295,7 +295,7 @@ module.exports = kind(
 	*/
 	updateKeyframes: function () {
 		if (this.useCssAnimation) {
-			this.$.spriteImage.set('stylesheetContent', this._generateKeyframes());
+			this.$.spriteImage.set('stylesheetContent', this._generateKeyframes('-webkit-') + this._generateKeyframes('-moz-') + this._generateKeyframes());
 		} else {
 			this._generatePositionList();
 		}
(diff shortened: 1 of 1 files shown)


src/SpriteAnimation/SpriteAnimation.js
```diff
@@ -493,8 +502,14 @@ module.exports = kind(
 	/**
 	* @private
 	*/
-	_generateKeyframe: function (percent, x, y) {
-		return (Math.ceil(percent*1000000) / 10000) +'%	{ -webkit-transform: translate3d('+ x +'px, '+ y +'px, 0);	transform: translate3d('+ x +'px, '+ y +'px, 0); }\n';
+	_generateKeyframe: function (percent, x, y, prefix) {
```
webOS101: Why not just change this method to generate all the prefixed methods at once and only call it once?  Or would that cause too big a change?

Djspaceg: That was actually how the first version of this PR looked, but it's unnecessary and wasteful to have -webkit- rules inside a @-moz-keyframes block, and vice versa. We also need standards-only blocks generatable for minimum rulesets. We Also need -webkit- prefixed block without -webkit- prefixed rules for the dozen versions of chrome that DO support standard transform but DON'T support @keyframe rule.

This creates the most efficient and minimally required source code.

src/SpriteAnimation/SpriteAnimation.js
```diff
@@ -493,8 +502,14 @@ module.exports = kind(
...
+		var outStr = (Math.ceil(percent*1000000) / 10000) +'%	{ ';
+		if (prefix) {
+			// add the prefix and the standard
+			outStr+= prefix + 'transform: translate3d('+ x +'px, '+ y +'px, 0); ';
```
aarontam: What do you think about throwing `'transform: translate3d('+ x +'px, '+ y +'px, 0); ` into a var so we don't have to repeat the concatenation?

Djspaceg: Yeah, that seems fine.

PR #1406, opened 2016-03-07
ENYO-3164: Make Control be the default kind.

aarontam:
### Issue
If `enyo/Control` is never required in an application, the default kind becomes `enyo/Component` instead.

### Fix
We explicitly require `enyo/Control` in `enyo/ViewController`.

### DIFF
--- a/src/ViewController.js
+++ b/src/ViewController.js
@@ -5,6 +5,9 @@ require('enyo');
 * @module enyo/ViewController
 */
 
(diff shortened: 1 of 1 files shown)


ryanjduffy:
Think this should be implemented in ViewController which already states that it works on Controls in its doc (and requires something that is at least API-compatible for things like `render()`) but acquires it implicitly via `kind.getDefaultCtor()`.

aarontam:
Yep that makes more sense - updated!

PR #1408, opened 2016-03-08
ENYO-3184: Remove srcElement property for generated media events.

aarontam:
### Issue
We had previously been providing a `srcElement` property for our generated media events. This is more of an optional change, and we are not utilizing this property anywhere internally, but in the spirit of being more standards-compliant and consistent, we should move away from this.

### Fix
We have removed the `srcElement` property when generating media events.


webOS101:
That's not deprecation, that's deletion! :)

aarontam:
Check again!

PR #1410, opened 2016-03-09
ENYO-3138: Recalculate scroll thumb metrics on resize

KunmyonChoi:
Issue:
NewThumb is created from scratch to support new smart share use case. Now we have new use case of resize the new data list at runtime. And, current version of scroll thumb doesn't include that logic.

Fix:
Recalculate metrics on resize handler. So that it can draw itself in correct size and position.

Enyo-DCO-1.1-Signed-Off-By: Kunmyon Choi (kunmyon.choi@lge.com)

### DIFF
--- a/src/NewThumb/NewThumb.js
+++ b/src/NewThumb/NewThumb.js
@@ -34,12 +34,12 @@ module.exports = kind(
 	* @private
 	*/
 	name: 'enyo.NewScrollThumb',
-	
+
 	kind: Control,
 
 	/**
 	* The orientation of the scroll indicator bar; 'v' for vertical or 'h' for horizontal.
-	* 
+	*
 	* @type {String}
 	* @default 'v'
 	* @public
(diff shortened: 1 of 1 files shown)


src/NewThumb/NewThumb.js
```diff
@@ -231,6 +231,14 @@ module.exports = kind(
 	},
 
 	/**
+	* Override `handleResize()` to re-calculate ratio and size.
+	* @private
+	*/
+	handleResize: function () {
+		this.calculateMetrics();
```
aarontam: Do we need to call the super?

KunmyonChoi: Generally speaking, yes it is always better to call super if there it is. :)
But, when I looked up the original behavior of UIComponent.handleResize, it is calling reflow of layout.
I thought it will be better to short circuit in this particular case because thumb is not using any layout kind.

PR #1412, opened 2016-03-11
PLAT-17000: Apply rAF into Control.applyStyle

ybsung:
Issue
------
: We need to reduce JS execution time to handle touchmove and to update DOM element while scrolling a list.
 
Fix
------
: I applied rAF into not enyo/ScrollMath.animate.fn() but enyo/Control.applyStyle() and called enyo/ScrollMath.animate.fn() directly when handing "touchmove" event.

Enyo-DCO-1.1-Signed-off-by: YB Sung <yb.sung@lge.com>

### DIFF
--- a/src/ScrollMath.js
+++ b/src/ScrollMath.js
@@ -19,7 +19,7 @@ var
 *
 * @event module:enyo/ScrollMath~ScrollMath#onScrollStart
 * @type {Object}
-* @property {Object} sender - The [component]{@link module:enyo/Component~Component} that most recently 
+* @property {Object} sender - The [component]{@link module:enyo/Component~Component} that most recently
 *	propagated the {@glossary event}.
 * @property {module:enyo/Scroller~Scroller~ScrollEvent} event - An [object]{@glossary Object} containing
 *	event information.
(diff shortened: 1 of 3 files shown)


src/ScrollMath.js
```diff
@@ -464,17 +462,14 @@ module.exports = kind(
 			y0 = this.y;
 			x0 = this.x;
 		});
-		this.job = animation.requestAnimationFrame(fn);
+		fn();
```
webOS101: If we're going to invoke this immediately why don't we save a bindSafely call and just remove the wrapping function?

src/touch.js
```diff
@@ -8,7 +8,10 @@ var
 
 var
 	Dom = require('./dom'),
-	Job = require('./job');
+	Job = require('./job'),
+	Control = require('./Control');
+
+var rAF = global.requestAnimationFrame;
```
webOS101: We have a wrapper around rAF to prevent issues where it doesn't exist on the platform:  animation.requestAnimationFrame.

ybsung:
Thanks, @webOS101.

I fixed the following issues that @webOS101 commented.
- Inserted the code in fn() into animate()
- Used animation.requestAnimationFrame instead of using window.requestAnimationFrame directly

And I merged the latest code in garnet-perf-exp into this PR.

0x64:
After we test this experimental branch, we must refine code.

PR #1414, opened 2016-03-14
ENYO-3222: Adding support for cancelAnimationFrame.

aarontam:
### Issue
Web engines such as Mozilla do not support `cancelRequestAnimationFrame`, and instead support the more-standard `cancelAnimationFrame`.

### Fix
Aside from some minor refactoring, we have provided support for either `cancelRequestAnimationFrame` or `cancelAnimationFrame` as older WebKit only supported `cancelRequestAnimationFrame`. Additionally, we have updated our animation API to provide a `cancelAnimationFrame` method and have deprecated the previous `cancelRequestAnimationFrame` method, in the name of consistency with the standard animation API. Also, considered having the deprecated `cancelRequestAnimationFrame` method call `cancelAnimationFrame`, but decided against it as the method effectively calls a single method itself.

### DIFF
--- a/src/animation.js
+++ b/src/animation.js
@@ -9,18 +9,21 @@ var
 	platform = require('./platform'),
 	utils = require('./utils');
 
-var ms = Math.round(1000/60);
-var prefix = ['webkit', 'moz', 'ms', 'o', ''];
-var r = 'requestAnimationFrame';
-var c = 'cancel' + utils.cap(r);
+var ms = Math.round(1000/60),
+	prefix = ['', 'webkit', 'moz', 'ms', 'o'],
+	rAF = 'requestAnimationFrame',
+	cRAF = 'cancelRequestAnimationFrame',
+	cAF = 'cancelAnimationFrame',
+	i, pl, p, wcRAF, wrAF, wcAF,
+	_requestFrame, _cancelFrame, cancelFrame;
 
 /*
 * Fallback on setTimeout
 *
 * @private
 */
-var _requestFrame = function(inCallback) {
-	return global.setTimeout(inCallback, ms);
+_requestFrame = function(callback) {
+	return global.setTimeout(callback, ms);
 };
 
 /*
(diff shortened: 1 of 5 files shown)


src/animation.js
```diff
@@ -28,23 +31,26 @@ var _requestFrame = function(inCallback) {
...
 * @private
 */
-var _cancelFrame = function(inId) {
-	return global.clearTimeout(inId);
+_cancelFrame = function(id) {
+	return global.clearTimeout(id);
 };
 
-for (var i = 0, pl = prefix.length, p, wc, wr; (p = prefix[i]) || i < pl; i++) {
+for (i = 0, pl = prefix.length, p, wcRAF, wrAF, wcAF; (p = prefix[i]) || i < pl; i++) {
```
ryanjduffy: There's no need for p, wcRAF, wrAF, or wcAF here since they were declared above.

src/animation.js
```diff
@@ -9,18 +9,21 @@ var
...
 	utils = require('./utils');
 
-var ms = Math.round(1000/60);
-var prefix = ['webkit', 'moz', 'ms', 'o', ''];
-var r = 'requestAnimationFrame';
-var c = 'cancel' + utils.cap(r);
+var ms = Math.round(1000/60),
+	prefix = ['', 'webkit', 'moz', 'ms', 'o'],
+	rAF = 'requestAnimationFrame',
+	cRAF = 'cancel' + utils.cap(rAF),
```
ryanjduffy: I realize this was existing code but it's pretty silly to add a function call so we don't have to type requestAnimationFrame again ...

aarontam: Agreed! Updated.

src/animation.js
aarontam: Boo JSHint - let's switch to ESLint...

src/animation.js
```diff
@@ -76,10 +82,21 @@ exports.requestAnimationFrame = function(callback, node) {
 /**
 * Cancels a requested animation callback with the specified id.
 *
+* @param {Number} id - The identifier of an animation request we wish to cancel.
+* @deprecated since 2.7.0
+* @public
+*/
+exports.cancelRequestAnimationFrame = function(id) {
```
ryanjduffy: Thoughts on removing the wrapping function here and just exporting `cancelFrame`? Same for the `cancelAnimationFrame` export below.

aarontam: Seems reasonable enough - should we do the same for `requestAnimationFrame` as well?

aarontam: Actually this won't work if we're using the native rAF and cAF methods as the scope will be incorrect. Vote to leave it as-is.

PR #1415, opened 2016-03-14
Fixes ENYO-3204 adds guard for setting the value of a select box in <= IE11

ryanjduffy:

### DIFF
--- a/src/Select.js
+++ b/src/Select.js
@@ -164,6 +164,25 @@ module.exports = kind(
 	},
 
 	/**
+	* Updates the `selected` option matching the current value. If the option doesn't have a value,
+	* it will match its content instead.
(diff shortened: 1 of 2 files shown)


src/Select.js
```diff
@@ -175,8 +194,15 @@ module.exports = kind(
 	* @private
 	*/
 	valueChanged: function () {
-		this.setNodeProperty('value', this.value);
-		this.updateSelectedFromNode();
+		// IE will clear selectedIndex when setting the value directly on the node so instead we
+		// iterate the children, find the matching value, and update selected if necessary.
+		if (platform.ie) {
```
viodragon2: we should check `platform.ie || platform.edge`

PR #1417, opened 2016-03-15
PLAT-17001: Not to change target of touchmove events

MikyungKim:
https://jira2.lgsvl.com/browse/PLAT-17001
Enyo-DCO-1.1-Signed-off-by: Mikyung Kim <mikyung27.kim@lge.com>


0x64:
Note: this patch assumes that we don't need to handle drag&drop or other special cases beside scrolling.

PR #1421, opened 2016-03-18
ENYO-3230: Do not bubble up onActivate event on rendered

viodragon2:
Issue
---
For every `active` checkboxes, it fires `onActivate` event on rendered. This can cause a problem as certain controls handles `onActivate` and may result in unexpected behaviors.

Fix
---
Do no bubble up `onAcivate` event on rendered.

### DIFF
--- a/src/Checkbox/Checkbox.js
+++ b/src/Checkbox/Checkbox.js
@@ -107,9 +107,6 @@ module.exports = kind(
 	rendered: kind.inherit(function (sup) {
 		return function() {
 			sup.apply(this, arguments);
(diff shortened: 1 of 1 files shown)


germboy:
While I generally agree with this approach, the group api `active` property is a finicky one. I'm slightly concerned with how this may affect existing apps that currently look for `onActivate` to fire on rendered, since this has been an expected behavior. I'll have to look back but I seem to remember certain apps doing specific things during onActivate as an application/view renders, based on Checkbox values.

PR #1422, opened 2016-03-21
ENYO-3219: Refresh scroll thresholds when adding / removing models

graynorton:
If a list is scrolled to the end and the size of the collection
changes, we need to recalculate the scroll thresholds and possibly
adjust the range of items to be rendered.

Until now, we've only adjusted thresholds in the `scroll()` method,
based on the direction and distance we've scrolled. In this fix, we
add a new method, `refreshThresholds()`, which recalculates the
thresholds based on the current scroll value. We call the new
method from `modelsAdded()` and `modelsRemoved()` to make sure that
the thresholds and item range are correct after the size of the
collection has changed.

### DIFF
--- a/src/NewDataList.js
+++ b/src/NewDataList.js
@@ -166,6 +166,7 @@ module.exports = kind(
 		// as long as calculateMetrics() is called only by reset().
 		this.numItems = num;
 	},
+
 	/**
 	* @private
 	*/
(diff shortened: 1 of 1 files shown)


src/NewDataList.js
```diff
@@ -209,6 +210,34 @@ module.exports = kind(
...
+	*/
+	refreshThresholds: function () {
+		var tt = this.threshold;
+
+		if (tt) {
+			var
+				v = (this.direction === 'vertical'),
+				val = v ? this.scrollTop : this.scrollLeft,
+				delta = this.delta,
+				cb = this.cachedBounds ? this.cachedBounds : this._getScrollBounds(),
```
KunmyonChoi: Does 'scroll threshold' also updated on showing change if we update collection while data list is hidden?

graynorton: @KunmyonChoi, at a glance, I don't think so. I think it would be relatively straightforward to make it happen, though -- we would just override `showingChangedHandler()` and call `refreshThresholds()` before calling the super method. If you write up a JIRA demonstrating a problem case, we will take a look.

PR #1426, opened 2016-03-22
ENYO-3244: Use strict equality check for content.

aarontam:
### Issue
For cases where we are changing the content from one falsy value to another (i.e. from the empty string `''` to `0`), the content does not update due to the equality check.

### Fix
I could not think of any cases where we would want to guard against updating the content when switching from one falsy value to another, so the guard has been changed to a strict equality check.

### DIFF
--- a/src/Control/Control.js
+++ b/src/Control/Control.js
@@ -1516,7 +1516,7 @@ var Control = module.exports = kind(
 		var was = this.content;
 		this.content = content;
 
-		if (was != content) this.notify('content', was, content);
+		if (was !== content) this.notify('content', was, content);
 
 		return this;
 	},


src/Control/Control.js
```diff
@@ -1516,7 +1516,7 @@ var Control = module.exports = kind(
 		var was = this.content;
 		this.content = content;
 
-		if (was != content) this.notify('content', was, content);
+		if (was !== content) this.notify('content', was, content);
```
webOS101: I think Cole's argument against this was that we should not be setting non-string values into content so we would be hiding developer error.

aarontam: (notimpressed)

Hm, so the alternative, in this situation, would be to explicitly convert `ProgressBar` popup content to strings? That seems a bit messy to me, but I can see the basis for the argument - `content` is specifically defined to be a string. Thoughts?

aarontam: This is actually located in the deprecated `setContent`, which I didn't notice during the first pass. When using the generic setter, there is no such restriction. I still think it might be worth updating this guard, at least for the sake of consistency.

webOS101:
Going to close this one as WNF.

PR #1435, opened 2016-04-07
ENYO-3248: Fix DataGridList resize issue

germboy:
### Issue

DataGridList does not always display list items after a resize.
### Fix

Clear the grid metrics when needed - also added method to adjust scroll position after a resize, so the top-most record index stays in the viewport.

Enyo-DCO-1.1-Signed-off-by: Jeremy Thomas jeremy.thomas@lge.com


lishammel:

PR #1437, opened 2016-04-18
ENYO-3219: Contrain values produced by `refreshThresholds()`

graynorton:
The recently added `refreshThresholds()` method was not properly
constraining the calculated threshold values in some cases. In this
change, we fix that oversight.

At the same time, we refactor `scroll()` to replace its own
threshold logic with the logic in `refreshThresholds()`, since the
old logic has become redundant.


lishammel:

PR #1440, opened 2016-04-20
ENYO-3333: apply opacity 0 on off screen light panel

KunmyonChoi:
Issue:
LightPanel is set visibility: hidden; on offscreen panel.
When element like checkbox is have visibility: visible; inside of
panel, then it will be shown even if panel is hidden. In popOnBack
case, check mark shown at the end of transition shortly. But, in
static panels case the check mark will be shown permenently.

Fix:
We apply opacity: 0; also to effectively hide all the children and
also not loosing the ability of blocking event and tap index handle.

Enyo-DCO-1.1-Signed-off-by: Kunmyon Choi (kunmyon.choi@lge.com)

### DIFF
--- a/src/LightPanels/LightPanels.less
+++ b/src/LightPanels/LightPanels.less
@@ -95,6 +95,7 @@
 
 		> .offscreen {
 			visibility: hidden;
+			opacity: 0;
 		}
 
 		.enyo-locale-right-to-left & > * {


src/LightPanels/LightPanels.less
```diff
@@ -95,6 +95,7 @@
 
 		> .offscreen {
 			visibility: hidden;
+			opacity: 0;
```
yeram: I thought that panel will be destroyed after off-screened, do we still need to have visibility: hidden prop?

aarontam: Hi @yeram, this will not always be the case i.e. if `popOnBack: false` is set and/or we are not caching views.

yeram: Aha! I missed that case. Thanks for informing!
But still have question.. Isn't it enough to just replace visibility: hidden to opacity: 0? I'm just curious...
If so we can remove visibility style. I referred table in this link.
http://stackoverflow.com/questions/272360/does-opacity0-have-exactly-the-same-effect-as-visibilityhidden

aarontam: Hi @yeram, no problem! If we do not set `visibility: hidden`, the panel will still respond to events and affect tab ordering.

PR #1444, opened 2016-05-06
Initial commit

anishramesan:
- Changes based on review
- Animationsupport an integral part
- Animate API for simple animations

Enyo-DCO-1.1-Signed-off-by: Anish Ramesan<anish.ramesan@lge.com>


anishramesan:
This PR is not required any more as the newer implementation will be pushed in different branch

PR #1451, opened 2016-06-12
PLAT-21885-Fix touch issue with Use Handle

rajyavardhanp:
Issue: Use handle doesn't collapse irrespective of no.of touches
Cause: mousedown and touchstart both the events are triggered on tap.
Fix: In the touchstart gesture event trigger mechanism, default event is
prevented as it is unnecessary for touch scenario.
Enyo-DCO-1.1-Signed-off-by: Rajyavardhan Reddy P
rajyavardhan.p@lge.com

### DIFF
--- a/src/gesture/touchGestures.js
+++ b/src/gesture/touchGestures.js
@@ -57,6 +57,7 @@ module.exports = {
 			var g = this.makeGesture('gesturestart', e, {vector: this.gesture, scale: 1, rotation: 0});
 			dispatcher.dispatch(g);
 		}
+		e.preventDefault();
 	},
 
 	/**


src/gesture/touchGestures.js
```diff
@@ -57,6 +57,7 @@ module.exports = {
 			var g = this.makeGesture('gesturestart', e, {vector: this.gesture, scale: 1, rotation: 0});
 			dispatcher.dispatch(g);
 		}
+		e.preventDefault();
```
aarontam: Won't this prevent native page scrolling on touch devices (only ones that support gestures, as we're making the change here instead of in touch.js)?

rajyavardhanp: Yes @aarontam , it creates problem in scrollers, I was working on the fix for that. As of now its not a proper fix. But I felt working with mouse events in touch mode was not proper behavior. Anyway I 'll  try to find a different solution.

lishammel:
Closing PR as Jira ticket is closed

PR #1455, opened 2016-07-06
PLAT-23332: Added orientation support to Resolution Independence

Djspaceg:
Enyo-DCO-1.1-Signed-off-by: Blake Stephens <blake.stephens@lge.com>

PR #1456, opened 2016-07-06
PLAT-23335: Added proportional scaling support to Resolution Independence

Djspaceg:
This change includes:
* PLAT-23332 #1455 
* Dynamic application of the base font size on the document object. No longer reliant on external CSS.
* A new enyo/dom method: removeBodyClasses
* RI configuration object
* Screen resize (or reorientation) event listeners for dynamic changes to the screen size
* Switches internal usage of string based current screen type to screenTypeObject for quicker lookups and ...
* More accurate body-class application with no overlap from previous executions.

### DIFF
--- a/src/resolution.js
+++ b/src/resolution.js
@@ -80,7 +80,7 @@ var ri = module.exports = {
 
 		if (rez.height > rez.width) {
 			portrait = true;
-			const swap = rez.width;
+			var swap = rez.width;
 			rez.width = rez.height;
 			rez.height = swap;
 		}
--- a/src/resolution.js
+++ b/src/resolution.js
@@ -1,14 +1,21 @@
 require('enyo');
 
 var
+	util = require('./utils'),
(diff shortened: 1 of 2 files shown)


src/resolution.js
```diff
@@ -25,6 +32,8 @@ var getScreenTypeObject = function (type) {
 * @module enyo/resolution
 */
 var ri = module.exports = {
+	config: {},
```
aarontam: Do we need this if we're assigning `config` to the default config object, further below?

Djspaceg: I guess not. But I figured this may need to be present if the RI object was referenced separately from the initialization.

aarontam: Fair enough - let's leave this in case the config init code changes or is removed, later.

aarontam:
Covered by https://github.com/enyojs/enyo/pull/1461.

PR #1458, opened 2016-07-14
ENYO-3411: Clear job function in case of destroy

yeram:
Job should be executed although control is destroy.

Enyo-DCO-1.1-Signed-off-by: Yeram Choi (yeram.choi@lge.com)


KunmyonChoi:
LGTM

PR #1472, opened 2017-01-02
[ENYO-3770]NewDataList: threshold is not updated between -Infinity and Infinity

Sangwook-Lee:
### Issue
In Photo&Video application use case, I can see the blank in the last of NewDataList.
Bacause thereshold's max is Infinity. So, threshold is not updated after scrolling down.
If we change from list that min of threshold is -Infinity and max of threashold is Infinity to other list that need to refresh threshold. At that time, max value of threshold is Infinity. This value have to be proper max value.

### Fix
I added `calculateMetrics` inside collection `ResetHandler` function in enyo.NewDataList.
This is for update when collection is changed by filter of bucketFilter.

ENYO-3770 NewDataList: threshold is not updated between -Infinity and Infinity
Enyo-DCO-1.1-Signed-off-by: Sangwook Lee sangwook1203.lee@lge.com

### DIFF
--- a/src/NewDataList.js
+++ b/src/NewDataList.js
@@ -30,7 +30,7 @@ module.exports = kind({
 	/**
 	* Returns an array of list items that are currently visible (whether partially
 	* or fully).
-	* 
+	*
 	* Experimental API -- subject to change.
 	*
 	* @public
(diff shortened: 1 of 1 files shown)


lishammel:

PR #1474, opened 2017-08-18
PLAT-43853: Number input field allows any character in Safari

anishlg:
### Issue Resolved / Feature Added
Enyo Input : Input allows any character in Safari

### Resolution
Safari not handled the same. regExp validation is done in keypress event

### Links
PLAT-43853

### Comments
Enyo-DCO-1.1-Signed-off-by: Anish TD(anish.td@lge.com)

### DIFF
--- a/src/Input/Input.js
+++ b/src/Input/Input.js
@@ -265,7 +265,7 @@ module.exports = kind(
 	* Fix to safari input type number allowing other characters
 	*/
 	testNumber: function(sender, e) {
-		if (platform.safari && this.type == 'number' && !/[eE.0-9+-]/.test(e.key)) {
+		if (platform.safari && this.type === 'number' && !/[eE.0-9+-]/.test(e.key)) {
 			e.preventDefault();
 		}
 	},
--- a/src/Input/Input.js
+++ b/src/Input/Input.js
@@ -265,7 +265,7 @@ module.exports = kind(
 	* Fix to safari input type number allowing other characters
 	*/
 	testNumber: function(sender, e) {
-		if (this.type == 'number' && !/[eE.0-9+-]/.test(e.key)) {
+		if (platform.safari && this.type == 'number' && !/[eE.0-9+-]/.test(e.key)) {
 			e.preventDefault();
 		}
 	},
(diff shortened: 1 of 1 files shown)


src/Input/Input.js
```diff
@@ -262,6 +263,15 @@ module.exports = kind(
 	/**
 	* @private
 	*/
+	testNumber: function(sender, e) {
+		if(this.type == 'number' && !/[eE.0-9+-]/.test(e.key)){ //&& !/[+\-]?(?:0|[1-9]\d*)(?:\.\d*)?(?:[eE][+\-]?\d+)?/.test(e.key)) {
```
sriramas: Remove the commented code.

sriramas:
LGTM!

src/Input/Input.js
```diff
@@ -167,7 +167,8 @@ module.exports = kind(
 		onfocus: 'focused',
 		oninput: 'input',
 		onclear: 'clear',
-		ondragstart: 'dragstart'
+		ondragstart: 'dragstart',
+		onkeypress: 'testNumber'
```
aarontam: We don't want this running on every keypress. I'm skeptical about even fixing this as it is a browser-level issue with Safari. At most, we should only add this handler for Safari.

src/Input/Input.js
```diff
@@ -261,6 +262,16 @@ module.exports = kind(
 
 	/**
 	* @private
+	* Fix to safari input type number allowing other characters
+	*/
+	testNumber: function(sender, e) {
+		if (this.type == 'number' && !/[eE.0-9+-]/.test(e.key)) {
```
aarontam: Let's use `===` here and place the regex outside of this statement and at the module level, so that it can be compiled ahead of time instead of created on each function call.

webOS101:
Why don't you just use the 'pattern' attribute?

aarontam:
Can we discriminate when the pattern is evaluated, though?

aarontam:
So we would want to apply the numeric pattern only for `number` input type, and if Safari, and only if no `pattern` is specified by the developer.

anishlg:
Numeric pattern may not be the ideal solution for this issue. pattern mainly for form validations and pattern will allow characters to input and validate on submit. this scenario not suited in this case. so added extra check for safari is done.
@aarontam , if you think this is not be fixed then i will close this issues

lishammel:

PR #1476, opened 2017-11-03
PLAT-47295: Resolution now better matches the screen shape

Djspaceg:
And measures `document.body` instead of `window` so components are scaled based on Enyo's available rendering area rather than the physical window dimensions.
Height is also factored into the check now too.

### DIFF
--- a/src/resolution.js
+++ b/src/resolution.js
@@ -9,7 +9,18 @@ var _baseScreen,
 	_orientation,
 	_riRatio,
 	_screenType,
-	_screenTypes = [ {name: 'standard', pxPerRem: 16, width: global.innerWidth,  height: global.innerHeight, aspectRatioName: 'standard', base: true} ],	// Assign one sane value in case defineScreenTypes is never run.
+	_workspaceBounds = {
+		width: (typeof global === 'object') ? global.innerWidth : 1920,
+		height: (typeof global === 'object') ? global.innerHeight : 1080
+	},
+	_screenTypes = [{
+		name: 'standard',
+		pxPerRem: 16,
+		width: _workspaceBounds.width,
+		height: _workspaceBounds.height,
+		aspectRatioName: 'standard',
+		base: true
+	}],	// Assign one sane type in case defineScreenTypes is never run.
 	_screenTypeObject,
 	_oldOrientation,
 	_oldScreenTypeObject,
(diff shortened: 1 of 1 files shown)


src/resolution.js
```diff
@@ -70,6 +81,25 @@ var ri = module.exports = {
 	},
 
 	/**
+	* Update the common measured boundary object. This object is used as "what size screen are we
+	* looking at". Providing no arguments has no effect and updates nothing.
+	*
+	* @memberOf ui/resolution
```
webOS101: Can you update this module to match the Enyo style

src/resolution.js
```diff
@@ -79,32 +109,31 @@ var ri = module.exports = {
 	* @public
 	*/
 	getScreenType: function (rez) {
-		rez = rez || {
-			height: global.innerHeight,
-			width: global.innerWidth
+		rez = rez || _workspaceBounds || {
```
webOS101: Is there something that clears `_workspaceBounds` so that the fallback would be needed?

src/resolution.js
```diff
@@ -79,32 +109,31 @@ var ri = module.exports = {
...
 			rez.height = swap;
 		}
 
-		// loop thorugh resolutions
-		for (i = types.length - 1; i >= 0; i--) {
-			// find the one that matches our current size or is smaller. default to the first.
-			if (rez.width <= types[i].width) {
+		// Loop thorugh resolutions, last->first, largest->smallest
+		for (var i = types.length - 1; i >= 0; i--) {
+			// Find the screenType that matches our current size or is smaller. Default to the first.
```
webOS101: Should this comment say '... matches our current size or is larger'?  I find 'default to first' to be a tad confusing because it's really the last in the list.  Maybe, default to largest?

src/resolution.js
Djspaceg: I don't believe there is anything that clears it, so we can probably remove this extra security.

src/resolution.js
Djspaceg: Just remove this line, right? I had removed it in a previous commit and I had to redo this a few times, and I must have forgotten about this in the latest attempt.

src/resolution.js
Djspaceg: I forgot that I mentioned "first" here. It was a previous iteration of the code. It's supposed to be "last", not "first" now. I can change it to "largest", but "last" is probably more accurate, considering the list is manually sorted, not automatically sorted, so the developer can choose the fallback order.

webOS101: What about the 'larger' comment?  It picks the screen type that is larger than (can contain) the current resolution, not one that is smaller than the current resolution.

Djspaceg: It's not checking for a larger size, it's checking for a smaller size. This part of the comment or the check didn't change from the previous version.

src/resolution.js
```diff
@@ -327,13 +352,13 @@ var ri = module.exports = {
 };
 
 ri.config = util.clone(configDefaults);
-ri.init();
+ri.init(document.body);
```
ryanjduffy: This should use `{measurementNode: document.body}` as below, right?

ryanjduffy: Also noticed (in strawman, at least) that this code runs before `body` has been created so both the current code and my suggested replacement are equivalent and likely wrong. Do we need to wrap this in `ready()` to wait for a valid node reference?

Djspaceg: Yes, you're probably right @ryanjduffy. It will need to run after the dom-ready event has fired, or at least the `body` tag is present.

